PR #3, opened 2016-02-12
Added kcoreaddons/src/lib/jobs to include_directories + updated from master branch

KitsuneRal:
Without this #includes<> inside kcoreaddons/ fail.


KitsuneRal:
Ok, that pull request got a bit more than a small build fix because I updated noKCoreAddons branch from 'master'. Nothing else than that.

Fxrh:
Hi kitsune,

Thanks for the pull request! As it is quite big, I'll need some time to look over it, which I unforntually don't have right now. But I will come back to it as soon as I have time.

Could you please do one pull request per feature in the future? Then it would be quite a bit easier for me to review them.

Fxrh:
Hi kitsune,

I merged now everything except the room names thing (which requires more time to look at).

KitsuneRal:
Thanks! Actually, most of this commit was simply merging a fresher master to noKCoreAddons. As for room names, I'll make a separate pull request (there's a room-names branch in my repo for that).

KitsuneRal:
Just noticed that you'd pulled the room names part as well - thanks!
Meanwhile, I start to think how to get rid of the noKCoreAddons branch - it's an unnecessary overhead to maintain it. I have 2 ideas so far:
1. Include the submodule from KDE into master and make a condition in CMakeLists.txt that would fallback to including and compiling the submodule if the system-wide KF5CoreAddons package wasn't found.
2. As far as I understand, you only use KJob out of KCoreAddons. Did you think of using QAsync instead? If there's no known problems with it, I think I'd try to migrate the code to pure Qt.

Fxrh:
As far as I can see I did not merge the room names part...

About the noKCoreAddons branch: Yes, I want to merge it back into master at some point. I'd prefer to make KCoreAddons an optional dependency and only include the submodule if it is not found.
If I'd have known that it is so complicated to get KCoreAddons running under Windows, I'd probably chosen not to use it and would have implemented it by myself... but I'm not quite sure what you mean with QAsync?

KitsuneRal:
Sorry, I meant Qt Concurrent, of course (for some reason I thought there's a Qt counterpart to std::async).

Fxrh:
Well, Qt Concurrent would be multithreaded, while KJob isn't (it just uses the Qt Event system), and as we are only waiting for IO in the jobs, I don't think we need multithreading.

If really needed, I would prefer to reimplement KJob, but as I don't like duplicated code, I'd like to somehow keep KCoreAddons.

KitsuneRal:
I'm afraid we're already in multithreading business. At least on Windows, I have the application stuck running after the main window closes. When I connected a debugger to it, I discovered about 10 threads running. Didn't look inside the call stacks though.

Fxrh:
The multithreading part should be the QML engine. That it is still running is normal, that can be changed with a flag somewhere when creating the main window (Qt::QuitAfterLastWindowClosed or something like that).

Fxrh:
So, now for the room name thing: All in all I really like it, thanks again for the pull request!

I would do a few things in the Room class different. Generally, the Room class should give you all information about a room (it's id, whether it has aliases, a name etc.). Additionally, it has a convenience function displayName(), which returns a good default to show as string representation for the room. Therefore, I'd propose a few changes:
1. Don't initialize d->name with d->id, as it is something different, and only keep the given name in it
2. Add a public function name(), which returns d->name
3. Calculate the displayName in displayName() directly. I don't think it is that computational expensive and we don't need another variable (as d->name should not be used for that)

Finally, for readability reasons, I'd prefer if/else instead of "?: " except for really short and obvious cases.

What do you think?

KitsuneRal:
All fair enough, will work on it later today (still within noKCoreAddons branch).

Fxrh:
cool, thanks! :)

KitsuneRal:
Done; please consider.

Fxrh:
Very nice :) I cherry-picked your commits into master and merged them into the noKCoreAddons branch.

PR #4, opened 2016-02-13
Added Qt Quick Controls to the list of dependencies to build + more documentation on noKCoreAddons build path on Windows

KitsuneRal:
If building Qt from scratch, qtquickcontrols should be explicitly added to the emerge line, otherwise the QML magic doesn't work even though the application builds without errors.


Fxrh:
Thanks for working on this!

If I've seen it correctly, parts of this are also in the pull request #3. Was that on purpose?

KitsuneRal:
Kind of on purpose, yes. Basically, I wanted to push the same change to master and noKCoreAddons. Thanks for finding the way through this - I'm a bit new to all this pull requests stuff...

Fxrh:
No problem, I know git is quite far from being user friendly ;)

PR #5, opened 2016-02-16
SyncJob::Private::parseEvents refactoring

KitsuneRal:
I took a refactoring run against repetitive code in `SyncJob::Private::parseEvents()` by making a poor-man's dispatcher loop. Besides, the whole SyncRoomData initialization seemed to belong to its own constructor, so I've made one. I guess the next place to refactor will be `Event::fromJson()` - it looks quite ugly and unscalable at the moment. Would you mind against a templated Event inheritor, like `RoomEvent<typename EvTraits>`?


KitsuneRal:
Looks like this has got too much cluttered with all the merges and stuff... I'll re-make a branch from your repo and replay the essential commit over there, for your convenience.

PR #6, opened 2016-02-22
SyncJob::Private::parseEvents refactoring

KitsuneRal:
I took a refactoring run against repetitive code in SyncJob::Private::parseEvents() by making a poor-man's dispatcher loop. Besides, the whole SyncRoomData initialization seemed to belong to its own constructor, so I've made one.


Fxrh:
Nice, thanks! And yes, it's quite a bit easier with no extra commits :)

PR #7, opened 2016-02-22
Process state changes in timeline events as well.

KitsuneRal:
This fixes display names of some rooms - 1:1 chats in particular.


Fxrh:
As discussed, this seems to be the right way to interpret the API ;) Thanks!

PR #8, opened 2016-02-23
Merging noKCoreAddons into master

KitsuneRal:
Basically it's adding up kcoreaddons the way they were added in the noKCoreAddons branch and then handcrafting conditional configuration and compilation in CMakeLists.txt.


Fxrh:
very good, I hope this makes working on quaternion a bit easier.

PR #12, opened 2016-02-28
Make Quaternion work with Qt 5.6

KitsuneRal:
Qt 5.6 introduced "HTTP redirects support" as they named it in What's new - thereby changing the default behaviour of QNetworkRequest: before 5.6 it used to follow redirects automatically, while in 5.6 you have to explicitly tell it to. Not sure if it's deliberate on their side - anyway, this commit makes use of the feature and fixes thumbnails loading that always goes through at least one redirect.


Fxrh:
Thanks!

PR #14, opened 2016-03-01
Fix avatar updates + minor cleanup

KitsuneRal:
For what it's worth, here's my call on the avatar updates stuff. It's basically the same but I thought it's worth it to factor out the connect/disconnect code so that once we have user data updates and other stuff coming around we could quickly add connects/disconnects to a single place (ok, two single places :) ).

I also accidentally found a messy dealing with User * in room.cpp, so I fixed that as well (it's a separate commit so you can cherrypick it if you don't like the main part of the PR).

### DIFF
--- a/client/models/userlistmodel.cpp
+++ b/client/models/userlistmodel.cpp
@@ -49,9 +49,9 @@ void UserListModel::setRoom(QMatrixClient::Room* room)
     {
         disconnect( m_currentRoom, &QMatrixClient::Room::userAdded, this, &UserListModel::userAdded );
         disconnect( m_currentRoom, &QMatrixClient::Room::userRemoved, this, &UserListModel::userRemoved );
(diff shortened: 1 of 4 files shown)


Fxrh:
About the NULL check: User* u will never be NULL, as the User is created if it doesn't exist yet (so the qDebug in the next line is unnecessairy).

And I don't quite see why we should move one function call into a new function. We can do that later, too, if more connections are needed later.

KitsuneRal:
The qDebug was in the original code, I just moved it. If we rely on a convention that Connection::user() always returns a valid user then the whole if{} construct is unnecessary.

Fxrh:
Yes, that's what I wanted to say. The library is partly still a mess, as I changed designs, but not al related code. Doing some major cleanup is on my TODO list.

PR #15, opened 2016-03-03
Add flexibility on choosing the bundled vs. system KCoreAddons.

KitsuneRal:
I've rehashed your changes in CMakeLists and made them work (at least for cases when KCoreAddons are not in the system - with or without BUILD_* option). Plus, I've made an additional option that would please package maintainers. This allows to provide the graceful fallback as a default behaviour, as it's in the mainline.
Sorry for the bulky commit - here's the list of changes:
1. Pump up required CMake version (it didn't reflect the actual contents).
2. Add FORCE_SYSTEM_KCOREADDONS so that Linux package maintainers (at least) could make sure they build against the system package, without graceful fallbacks.
3. Add respective logic around find_package(KF5CoreAddons). Enable the graceful fallback again as a default (all options are OFF).
4. Fix adding sources from kcoreaddons/* to SRCS (the previous condition wasn't right). If FORCE_SYSTEM_KCOREADDONS is set we won't even reach this place so I don't check it.
5. Set compiler options in a nicer and more readable way.
6. Former KCOREADDONS_FOUND preprocessor macro is renamed with USING_SYSTEM_KCOREADDONS - this name better reflects the purpose.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.11)
+cmake_minimum_required(VERSION 3.1) # Because of target_compile_features
 
 project(quaternion)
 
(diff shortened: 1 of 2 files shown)


Fxrh:
Very nice, thanks!

PR #16, opened 2016-03-04
Minor CMakeLists polishing

KitsuneRal:
1. Show the path to Qt (to double check in case of several Qt versions installed).
2. Show the path to KCoreAddons if BUNDLE_KCoreAddons is OFF (not only if AUTO).

I guess it's ready for master.


Fxrh:
Ok, I'll merge it, thanks!

PR #18, opened 2016-03-15
Sync on timer

KitsuneRal:
This is independent (as much as possible) from the pull request about user/room names and based on the same commit as the other one. Basically, I've made syncs work the way I think is proper - by a timer rather than immediately invoke the next sync once the previous is finished. The commits in this PR should be taken together - the last one fixes a linking error by the first one.


KitsuneRal:
This one is obsoleted by PR#19.

PR #23, opened 2016-05-06
Use lambda capture instead of throwing a job pointer around

KitsuneRal:
This is a counterpart of a (bigger) PR for the lib.


KitsuneRal:
Given the feedback in #quaternion:, I thought I'd (almost) revert this PR, proposing just a signal change instead. Truly, in this case using a lambda only confuses the flow.

Fxrh:
Thanks! :)

PR #25, opened 2016-06-01
Improved UX during login, cleanup of orphaned connections & select able room topics

andir:
These three commits change some UX issues and cleanup old connections from previous connect attempts.

The Login dialog should not just freeze while connecting. It should reflect the processing state and avoid further login from the user.

When logging in with invalid user/password/server the old connection was never cleaned up. At the same time I've added a new signal to bounce information about a new connection to all involved components. This will be more useful once I can add the StatusBar commit (requires the Connection::syncStarted signal in libqmatrixclient).

Finally the topic of a channel should be select able. People can now copy&paste it somewhere else and at least open URIs until there is proper support.


Fxrh:
Looks good, thanks!

PR #33, opened 2016-06-10
nullptr invasion

KitsuneRal:
This is the same as in libqmatrixclient some time ago - replacing 0's with nullptr's where appropriate.


Fxrh:
Thanks!

PR #34, opened 2016-06-13
Massive update to add libqmatrixclient to the picture

KitsuneRal:
Also deduplicated parts of README here and there. And the link to the instructions about building KDE on Windows have been updated; the whole section shifted to non-recommended status as it's really hard comparing to just picking Qt5+toolchain from qt.io and KCoreAddons from git.


Fxrh:
Thanks!

PR #35, opened 2016-06-14
Fixed a failing assert inside QList<> when syncing a room with no messages

KitsuneRal:
I caused this crash in a bit unfair way - by setting the number of timeline events in the filter down to zero. However, the case is valid: right now QuaternionRoom initializes unread messages flag with `true` even though there are no _any_ message events (yet). Probably there is a more profound fix that makes sure the flag and the container are consistent but this is the simplest fix that can be done in no time.


Fxrh:
Thanks!

PR #36, opened 2016-07-15
Added saving login credentials.

jklmnn:
I have added a saving option for login credentials.
Only problem would be the use of the password instead of the token.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
-build
+build*
 .kdev4
+*.user
--- a/client/logindialog.cpp
+++ b/client/logindialog.cpp
@@ -42,7 +42,7 @@ LoginDialog::LoginDialog(QWidget* parent)
     formLayout->addRow("Server", serverEdit);
     formLayout->addRow("User", userEdit);
     formLayout->addRow("Password", passwordEdit);
-    formLayout->addRow("Save login credentials", saveCheck);
+    formLayout->addRow("Stay logged in", saveCheck);
     
     QVBoxLayout* mainLayout = new QVBoxLayout();
     mainLayout->addLayout(formLayout);
@@ -54,8 +54,6 @@ LoginDialog::LoginDialog(QWidget* parent)
     settings = new QSettings(QString("Quaternion"));
     settings->beginGroup("Login");
     serverEdit->setText(settings->value("homeserver", "https://matrix.org").toString());
-    userEdit->setText(settings->value("username", "").toString());
-    passwordEdit->setText(settings->value("password", "").toString());
     saveCheck->setChecked(settings->value("savecredentials", false).toBool());
     settings->endGroup();
 
@@ -82,11 +80,6 @@ void LoginDialog::login()
     settings->setValue("savecredentials", saveCheck->isChecked());
     if(saveCheck->isChecked()){
         settings->setValue("homeserver", serverEdit->text());
-        settings->setValue("username", user);
-        settings->setValue("password", password);
-    }else{
-        settings->remove("username");
-        settings->remove("password");
     }
     settings->endGroup();
 
--- a/client/logindialog.cpp
+++ b/client/logindialog.cpp
@@ -36,11 +36,13 @@ LoginDialog::LoginDialog(QWidget* parent)
     passwordEdit->setEchoMode( QLineEdit::Password );
     sessionLabel = new QLabel("Session:");
     loginButton = new QPushButton("Login");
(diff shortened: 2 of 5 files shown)


jklmnn:
I have tried to use the token, but `m_connection->token()` and `m_connection->userId` both return empty strings. Otherwise just saving them and using `m_connection->connectWithToken()` would not be a problem.

Half-Shot:
:+1: Thanks, I realise it's a pain but I'd rather get it right from the go :)

jklmnn:
Before we merge, I'm currently creating a logout option (otherwise you would have to delete the conf file manually).

jklmnn:
I have added a logout option that acts the same as quit but removes userid and token from the settings. It is disabled if stay logged in was not chosen.

KitsuneRal:
Wait, let's try to fix the empty m_connection->token() in the first place. @jklmnn, at which point in time are you checking it? It's supposed to have a sensible value at the moment Connection emits the connected() signal. Ditto user_id.

KitsuneRal:
Now that I look closer into your code... You're actually saving credentials at the wrong moment. First, I don't see much value in storing the password (even besides the security concern) _before_ it's validated at the server. So I'd rather store at least the ~~password~~access token in the Connection::connected() handling slot (it is QDialog::accept() atm, and that will have to be changed). By that moment you'll be able to pick all the data from m_connection. Next, the server may return a different hostname and a different user_id from those passed to it - so you might consider to save ~~credentials~~ access data two times, first before (for the case when the user quickly closes the application before the login succeeds) and then after logging in.

Now that you don't have to mess with passwords, you might want to consider the case when the login dialog is skipped completely - once saving the access_token works, it will be obvious to start Quaternion already logged in, if access_token is valid - that's how Vector works, by the way.

davidar:
> Now that you don't have to mess with passwords, you might want to consider the case when the login dialog is skipped completely - once saving the access_token works, it will be obvious to start Quaternion already logged in, if access_token is valid - that's how Vector works, by the way.

That's also how tensor works, fwiw:

- https://github.com/davidar/tensor/blob/master/client/qml/Tensor.qml#L32-L41
- https://github.com/davidar/tensor/blob/master/client/qml/Tensor.qml#L93-L98

jklmnn:
@KitsuneRal I have tried to set it before `connected` was called, so I think that didn't work. Have you looked into the right commit? I'm not storing the password anymore, I have changed this in bccb8f7.
I save token and userid from the connection in `mainwindow.cpp` after the login dialog returned.

KitsuneRal:
Doh, I've just found a commit in the PR that does exactly what I described. Sorry for the mess, discard my previous comment :)

Now to the (hopefully) things that are not addressed yet.

- I wonder why you'd want to keep QSettings* as a 1) dynamically allocated; 2) member variable. I suppose it's only used in MainWindow::initialize() right now, and if it's used somewhere else, it's not a problem to make a local object in that other place.
- Duplicating the code that creates a connection looks a bit dirty. Along the way, LoginDialog::connectionChanged() signal has never looked right in the first place (the usual `dialog.exec()` completion suffices); it should be factored out.

Actually, I independently tried to implement logoutAction to test the libqmatrixclient code. I'm merging your code with mine now, taking the best parts from both implementations (unfortunately, not here because I only have push rights for libqmatrixclient).

jklmnn:
I had the idea that we could use the QSettings for more than just saving credentials. You also could save the window layout and maybe notification settings. So I thought it would make more sense to keep it as member object.

Since the connection changed was (somehow in my eyes) tied to LoginDialog, I kept it there and just copied the code inline without creating an new slot. I did not want to call LoginDialog at all since I just had to set token and userid.

KitsuneRal:
Something like that: https://github.com/KitsuneRal/Quaternion/tree/logout - but this needs further cleanup. And I've thrown a bit too many things into one commit (my bad).
Regarding QSettings - this is exactly the reason why I wouldn't make it a member of MainWindow. We already need it in LoginDialog, for example. And most likely will need in more places. So I guess we'd just make a QSettings object in the place we need it. And we don't need to pass the name to QSettings constructor - just set the application name once in main.cpp.

jklmnn:
Looks good to me, already a lot cleaner than what I have ;)
The usage of QSettings in different classes is a point.

KitsuneRal:
@jklmnn, many thanks again for initial work on this. You can look at #52 to find the PR under review. This one gets closed now.

PR #37, opened 2016-07-22
More guards against null Connection* members

KitsuneRal:
When a logout action is introduced, we won't be able to rely on the fact that a Connection always exists.

A bit unobvious change is in imageprovider.cpp. The idea is that we have to check m_connection immediately before using it, rather than before dispatching a method call through an event loop, which can take a considerable time and lose the pointer to Connection along the way.


Fxrh:
Looks good, thanks! :)

PR #38, opened 2016-07-26
Dropping KCoreAddons

KitsuneRal:
This matches Fxrh/libqmatrixclient#15. Note that these two PRs should go together, since I had to replace KJob with BaseJob throughout in all places where pointers to job objects are still used. Other than that, it's only adding a couple of #includes and updating CMakeFiles.txt / README.md.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "lib"]
 	path = lib
-	url = ../libqmatrixclient.git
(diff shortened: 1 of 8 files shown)


Fxrh:
hmm, there seems to be a problem in the lib...

KitsuneRal:
I fixed the build - this works now.

PR #41, opened 2016-08-25
More chat widget love

KitsuneRal:
An assorted set of tweaks and small fixes. On a visual side - images now fit the viewport and date sections now have background and stick to the top of the chat.
This does not enable rich text (yet) - a separate PR will come for that.


KitsuneRal:
Before this gets in - I got fed up with not very convenient tooltip for the message source and hacked up a context menu with a "View source" menu item. The tooltip is basically unused now but we might find a good use for it later on.

PR #43, opened 2016-08-27
Even more chat widget love

KitsuneRal:
A quickfix-style replacement for the ScrollView; and some more QML tweaks.

### DIFF
--- a/client/qml/chat.qml
+++ b/client/qml/chat.qml
@@ -11,61 +11,82 @@ Rectangle {
         chatView.positionViewAtEnd();
     }
 
-    ScrollView {
-    anchors.fill: parent
+    ListView {
+        id: chatView
+        anchors.fill: parent
 
-        ListView {
-            id: chatView
-            anchors.fill: parent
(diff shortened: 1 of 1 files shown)


KitsuneRal:
Terrible flicking FPS caused me do go for some profiling. Turned out that ScrollView in a TextArea tries to calculate something even when not shown; after moving it out to a Loader, the animation is again rather smooth. I also provide a bit of feedback on the slider even when flicking without its usage. This feedback looks a bit weird (although physically correct :) ) when bouncing on bounds, so I removed bouncing as well.

KitsuneRal:
Terrible flicking FPS caused me do go for some profiling. Turned out that ScrollView in a TextArea tries to calculate something even when not shown; after moving it out to a Loader, the animation is again rather smooth. I also provide a bit of feedback on the slider even when flicking without its usage. This feedback looks a bit weird (although physically correct :) ) when bouncing on bounds, so I removed bouncing as well.

PR #45, opened 2016-08-30
Rich text in timeline

KitsuneRal:
This relies on the previous PR(#43) - the new commits are actually just two most recent ones. Most prominently, this adds support of HTML messages (in particular, those produced by Vector). Because emotes can be HTML-based too, I changed the way they are rendered - it looks a bit unusual (definitely not weechat-like) so I really want your feedback on this. If you don't like it, I can do another commit to include the room member name into the content.
Also, as a bonus, comes proper rendering of topic events inside the timeline.


Fxrh:
Looks good :)

PR #49, opened 2016-09-01
Add .desktop file for Linux

Takios:


Takios:
Removed the MimeType because the protocol is not standardized.

Fxrh:
Thanks!

PR #52, opened 2016-09-06
Account settings and logout

KitsuneRal:
This brings in the following major things:
 - Storing account settings (homeserver, userid, access_token and keep_logged_in) using `QMatrixClient::AccountSettings`. This has been done with potential gaining of multiple accounts support in mind, so it's now a bit more generic than it needs for a single account. In particular, settings are stored under a separate group (keyed by a fully qualified userid) inside "Accounts" group. I also tried to write the code so that it would be easier to extend it on multiple accounts later on.
 - Automatic login upon starting the application, if access_token has been saved.
 - Logging out. This required removing all those race conditions due to syncJob's running independently almost at any point in time. Fortunately it came to be fairly easy, as soon as those syncJob's are explicitly abandoned. Active jobs prevent deleteLater() to actually delete an object, as it turned out - so a pure fire-and-forget at least for long-polling jobs is not quite workable, it seems. Note that, by debug logs, jobs still take their time to actually be destroyed; but since they are already disconnected and don't have active network connections underneath, they are pretty much benign.

### DIFF
--- a/client/chatroomwidget.cpp
+++ b/client/chatroomwidget.cpp
@@ -119,6 +119,7 @@ void ChatRoomWidget::setRoom(QMatrixClient::Room* room)
 
 void ChatRoomWidget::setConnection(QMatrixClient::Connection* connection)
 {
+    setRoom(nullptr);
     m_currentConnection = connection;
     m_imageProvider->setConnection(connection);
     m_messageModel->setConnection(connection);
--- a/client/logindialog.cpp
+++ b/client/logindialog.cpp
@@ -19,28 +19,35 @@
 
 #include "logindialog.h"
 
-#include <QtCore/QDebug>
-#include <QtCore/QUrl>
+#include <QtWidgets/QLineEdit>
+#include <QtWidgets/QPushButton>
+#include <QtWidgets/QLabel>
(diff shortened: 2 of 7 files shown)


KitsuneRal:
Note that this relies on a not-yet accepted Fxrh/libqmatrixclient#24 (but AppVeyor's failure is not due to that).

KitsuneRal:
Upon some reading - looks like VS2013's failure is not quite conforming to C++ standard (VS2015 manages to build the code without problems). Given that I don't have considerable interest in building with VS in general (I'm fine to use GCC and Clang even on Windows), a probable response will be switching to VS2015 and officially declaring that Quaternion is buildable only with that version from MS. But I'll try to dig a bit more.

KitsuneRal:
Finally :)

Fxrh:
Hmm, using this, all channels are shown for me to contain new messages... Are you able to reproduce that? (Maybe it's some change in libqmatrixclient that wasn't in Quaternion yet)

KitsuneRal:
Reproduced it, reverted the breaking commit (will look into it later) - highlighting should work again.

Fxrh:
Ship it! :)

Issue #31, opened 2016-06-06
Quaternion does not exit if window is closed


Fxrh:
Connection->Quit does work, though

KitsuneRal:
Seems to be fixed by dropping KCoreAddons - if not, please reopen.

Fxrh:
It seems Quaternion was only properly quit if qApp->quit() was called. At least for me, the problem still exists if I just close the window, so unfortunately I have to reopen this...

Fxrh:
I don't quite understand what's wrong here... I asked in the Qt forum: https://forum.qt.io/topic/71112/application-does-not-quit

Fxrh:
Might be a Qt bug. c39e1f6c33edcc18c1b20c0f94f10503d1bfd828 has a workaround for this.

PR #58, opened 2016-09-08
Save geometry of the main window and state of dock widgets

### LINKED ISSUE #55: Last window size should be remembered
Currently, everytime you launch the client, it's window size is reset to whatever is the default. It would be preferable to "remember" the old window size, and set it back to that on the next launch.

KitsuneRal:
Closes #55.
Also, menu creation has been moved out to a separate method and invoked before showing the window, not afterwards.

### DIFF
--- a/client/mainwindow.cpp
+++ b/client/mainwindow.cpp
@@ -75,7 +75,7 @@ void MainWindow::createMenu()
 
     auto quitAction = connectionMenu->addAction(tr("&Quit"));
     quitAction->setShortcut(QKeySequence::Quit);
-    connect( quitAction, &QAction::triggered, qApp, &QApplication::quit );
+    connect( quitAction, &QAction::triggered, qApp, &QApplication::closeAllWindows );
 
     // Room menu
     auto roomMenu = menuBar()->addMenu(tr("&Room"));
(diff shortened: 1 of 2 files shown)


client/mainwindow.cpp
```diff
@@ -210,7 +232,7 @@ void MainWindow::connectionError(QString error)
 void MainWindow::closeEvent(QCloseEvent* event)
 {
     setConnection(nullptr);
-
+    saveSettings();
```
Fxrh: I think MainWindow::closeEvent is not invoked when closed via the "Quit" action. Maybe we should use MainWindow::close there instead of QApplication::quit?

KitsuneRal: Good call. In fact, I thought that saving the state just on exit might fall short in, e.g., a case of a sudden crash after two days of working in a row - then upon restart Quaternion opens in a two-days-old configuration. Probably it makes sense to save state upon every resize or move - or its it too much of hustle?
At any rate, you're right: using `QApplication::quit()` for finishing is definitely not right, now that I read Qt docs (it essentially kills the event queue at once - very quick but very harsh). `QApplication::closeAllWindows()` seems to be most proper (it's a bit more universal than `QMainWindow::close()`).

Fxrh:
Hmm, now I have the problem again that Quaternion does not correctly close... Does it for you?

KitsuneRal:
Don't have any (just tested several times, by various means including Connection->Quit, and from various states - before login, after opening a timeline etc.). Might be specific to Linux.

Fxrh:
This should be fixed by c39e1f6, see issue #31. Feel free to merge :)

PR #59, opened 2016-09-09
Better status in login dialog

KitsuneRal:
A small usability improvement - see details in the second commit.

### DIFF
--- a/client/mainwindow.cpp
+++ b/client/mainwindow.cpp
@@ -69,7 +69,7 @@ void MainWindow::initialize()
         auto connectionMenu = menuBar->addMenu(tr("&Connection"));
 
         loginAction = connectionMenu->addAction(tr("&Login..."));
-        connect( loginAction, &QAction::triggered, this, &MainWindow::showLoginWindow );
+        connect( loginAction, &QAction::triggered, [=]{ showLoginWindow(); } );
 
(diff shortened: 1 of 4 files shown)


client/mainwindow.cpp
```diff
@@ -122,14 +122,15 @@ void MainWindow::setConnection(QuaternionConnection* newConnection)
         connect( connection, &Connection::syncDone, this, &MainWindow::gotEvents );
         connect( connection, &Connection::connected, this, &MainWindow::getNewEvents );
         connect( connection, &Connection::reconnected, this, &MainWindow::getNewEvents );
-        connect( connection, &Connection::loginError, this, &MainWindow::loggedOut);
-        connect( connection, &Connection::loggedOut, this, &MainWindow::loggedOut);
+        connect( connection, &Connection::loginError, this, &MainWindow::loggedOut );
+        connect( connection, &Connection::loggedOut, [=]{ loggedOut(); } );
```
Fxrh: Hmm, Signals&Slots is not intelligent enough to ignore arguments with default values?

Fxrh:
:+1:

client/mainwindow.cpp
KitsuneRal: Not so much, unfortunately. The good news is that the lambda syntax is even more compact than the original one :)

PR #60, opened 2016-09-09
Strictly require Qt libraries

KitsuneRal:
In particular, people tend to miss the warning about Qt Quick (qt-declarative5 on Debian derivations) not installed, and the build failure after that is not so specific.


Fxrh:
:+1: Looks good, thanks :)

PR #62, opened 2016-09-10
Show a progress bar during the initial sync

Fxrh:
Make sure the user sees something is happening during the initial sync


KitsuneRal:
The idea is interesting but I'd also supply some message(s) to the status bar, now that we have it. Such as "Connecting to <server>" and "Connected to <server> as <mxid>".
I also don't know for KDE but on Windows the progress bar looks a bit clumsy. But that may be just me.

Fxrh:
What do you mean with clumsy? On Linux, it looks good imho but it is a bit long (which could probably be solved by putting more functionality into the status bar.
To put in text messages, too, we'd need to put the progress bar into the permanent widgets, as temporary widgets and text messages share the same place.

Fxrh:
quaternion_progress_bar This is how it looks on Linux, btw

Fxrh:
Screenshot of the new version:
quaternion_progress_bar_2

PureTryOut:
That last version looks better, however I would initially give the progress bar the same width as the users list above it. However since that window can be closed or resized at any time, I'm not sure how that would work.

KitsuneRal:
If we are in for aligning widths then I'd stuff the progress bar in the bottom of the room list. A nice side effect in that case is that upon completion of the first sync it's the room list that gets filled up.

KitsuneRal:
As for how it looks on Windows...
progress-bar-on-windows

KitsuneRal:
Not sure if this (Windows looks) can be helped without aggressive themeing - which I'm not inclined to do anyway. Windows 10 might look better, by the way.

Fxrh:
Ok, it looks quite strange on Windows, especially because it's quite a bit to the right.

I'm not sure whether we can align it to the size of the users dock widget.

Putting it into the room list might work, I would prefer to keep it in the status bar, though. But if we can't let it look not strange in the status bar...

KitsuneRal:
I wonder if some kind of a work-in-progress icon could serve for this (given that we don't show progress steps). Or just disable the progress bar on Windows - I'm perfectly fine with just a message in the status bar, if that platform has so poor defaults for the progress bar outlooks :)

Fxrh:
I used a busy indicator now instead of a progress bar, hopefully this way it looks better on Windows. Additionally, I fixed the problem with the indicator/progress bar not being correctly aligned to the right side.

busy_indicator

KitsuneRal:
I've just checked it again, and I have to admit that on bigger screens the indicator gets lost somewhere in the bottom-right corner now and is not quickly associated with the short message in the bottom-left... Anyway, it's good enough for me at the moment; my personal dream would be to have a pop-up progress indicator somewhere in the bottom-right corner that would have both the message and the spinner/progress but this obviously waits for later. The only thing I would add to this, maybe, is a status line "Connected to <server> as <mxid>" instead of clearing the status bar completely.

PR #65, opened 2016-09-16
Fix messages ordering

KitsuneRal:
This accompanies Fxrh/libqmatrixclient#30. One notable thing not directly related to ordering is that `MessageEventModel` now uses the messages container of `QuaternionRoom` instead of maintaining its own copy.

### DIFF
--- a/client/models/messageeventmodel.cpp
+++ b/client/models/messageeventmodel.cpp
@@ -36,10 +36,9 @@
 
 MessageEventModel::MessageEventModel(QObject* parent)
     : QAbstractListModel(parent)
-{
-    m_currentRoom = nullptr;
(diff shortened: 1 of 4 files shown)


Fxrh:
hmm, looks like the history is added in the wrong order?

KitsuneRal:
Oh, right :-| Fixing it.

Also, upon reading the spec I realized that the current `Room::updateData` doesn't correctly process the case when a timeline message is also a state change message. E.g., all user displayname changes will occur before adding all messages from this sync to the timeline - which is not quite accurate. On the other hand, I'm seriously unsure that the current code properly deals with such cases overall. Needs some audit.

PR #68, opened 2016-09-20
Sort room member list

KitsuneRal:
That's a quick and simple implementation of room member list sorting. It boils down to a comparator class that is used every time when `UserListModel::m_users` is accessed. A more careful approach would be to imbue the `m_users`' type with self-sorting features but that implies much more keystrokes (though the same code will most likely be useful in other places of the application). I'll leave that kind of refactoring for another time (and apparently will make it in libqmatrixclient).
A note on `lowerBoundIndex`: I made it a template in order to unbind from a certain container type because `m_users` is a very good candidate for `QList`->`QVector` replacement. Needless to say it's only instantiated once.

### DIFF
--- a/client/models/userlistmodel.cpp
+++ b/client/models/userlistmodel.cpp
@@ -27,6 +27,36 @@
 #include "lib/room.h"
 #include "lib/user.h"
 
+class MemberNameSorter
+{
+        using Room = QMatrixClient::Room;
+        using User = QMatrixClient::User;
+    public:
+        MemberNameSorter(Room* r) : room(r) { }
+
+        bool operator()(User* u1, User* u2) const
+        {
+            auto n1 = room->roomMembername(u1);
+            auto n2 = room->roomMembername(u2);
+            if (n1[0] == '@')
+                n1.remove(0, 1);
+            if (n2[0] == '@')
+                n2.remove(0, 1);
+            return n1 < n2;
+        }
+
+        template <typename ContT>
+        typename ContT::size_type lowerBoundIndex(const ContT& c,
+                                                  typename ContT::value_type v) const
+        {
+            return  std::lower_bound(c.begin(), c.end(), v, *this) - c.begin();
+        }
+
+    private:
+        Room* room;
+};
+
+
 UserListModel::UserListModel(QObject* parent)
     : QAbstractListModel(parent)
 {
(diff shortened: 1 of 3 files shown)


client/models/userlistmodel.cpp
```diff
@@ -103,32 +135,41 @@ int UserListModel::rowCount(const QModelIndex& parent) const
...
+        disconnect( user, &QMatrixClient::User::avatarChanged, this, &UserListModel::avatarChanged );
+    } else
+        qWarning() << "Trying to remove a room member not in the user list";
 }
 
 void UserListModel::memberRenamed(QMatrixClient::User *user)
 {
-    int pos = m_users.indexOf(user);
-    if ( pos > -1 )
+    auto pos = MemberNameSorter(m_currentRoom).lowerBoundIndex(m_users, user);
```
Fxrh: Hmm, so we don't update the position in the list?

KitsuneRal: Ah, good catch! We should update it, obviously.

KitsuneRal: Turns out Quaternion doesn't really support renaming, to begin with :-D

Fxrh: Hmm, it would be nice if you'd fix it anyway now, as we don't need to remember this when implementing user name changes in the future then. But if you don't want to work on it without being able to test it, I'm ok with merging now, too :)

KitsuneRal: I guess not much is needed to make it working so I'll rather take time.

KitsuneRal: The good news is that in fact Quaternion does process renaming, only doesn't reflect it in the timeline properly. The bad news is that renaming affects the whole timeline because the user name is calculated for every line based on the _current_ user state, not historical state. Getting historical state is non-trivial, unless we resort to calling the server for a state per certain moment - it needs to un-play all state changes that happened after that certain moment until present time.

Given that currently renamings cause rather awful, though funny, changes in the timeline, I put this on hold, and my next hunting, apparently, will be for this thing. Needless to say, most of the addressing code will most likely land in the lib, not in the client code.

Fxrh: I think we do not need to show the fully distinguished name there, but just the name the user selected. We probably need the replaced state though, but this one is linked from the event (we would need an event lookup function for that, but this would also help for removing duplicate events)

As text something like "<id> changed their name from <oldname> to <newname>"?

Fxrh: Hmm, we might be missing the oldname, if we go back in history, though...

Fxrh: As discussed in the channel, room member events seem to have a prev_content field (below "unsigned")

KitsuneRal: Yes, actually I already found them there yesterday (at least in what matrix.org gave) but started thinking about a more overall solution - such as a sequence of state events (or even room states theirselves - to save on replaying/unplaying events every time `MessageEventModel::data` needs them to). The point is that we need state information not only for the timeline message about renaming but also for messages from that member before and after (we have to put different names in AuthorRole).

KitsuneRal:
As fixing the timeline is a bigger endeavour, I think I'll do the position update on renaming and request a merge. And the time machine for room states will come separately.

Fxrh:
Ok, merging now :) Thanks again!

KitsuneRal:
Uhm, actually I thought I'd first make it properly reorder on renaming... Ok, I'll do another PR for that.

Fxrh:
Oh, sorry, seems like I was already a bit tired...

KitsuneRal:
Actually, it turned to be for the good, because when I started digging that... In short, without room states time machine it's very complicated code. With it, it's much easier.

PR #70, opened 2016-09-23
Display tweaks

KitsuneRal:
Some highly visual improvements, so testing is very much welcome (weekend ahead :) ):
 - using sender to enhance status messages;
 - renames were in the previous commit, reflecting canonical aliases in the timeline now arrives;
 - most visual: instead of using specific colours I rewired both the timeline and the room list to something less intrusive into the desktop palette - users of dark themes are welcome.

The last change made Quaternion less colourful than it used to be; but I don't think this can be helped without breaking so colour schemas that can be so diverse. Feedback is much welcome though.

### DIFF
--- a/client/models/messageeventmodel.cpp
+++ b/client/models/messageeventmodel.cpp
@@ -20,6 +20,7 @@
 #include "messageeventmodel.h"
 
 #include <QtCore/QStringBuilder>
+#include <QtCore/QSettings>
 #include <QtCore/QDebug>
 
 #include "../message.h"
@@ -285,6 +286,14 @@ QVariant MessageEventModel::data(const QModelIndex& index, int role) const
     if( role == HighlightRole )
     {
         return message->highlight();
+    }
+
+    if( role == Qt::DecorationRole )
+    {
+        if (message->highlight())
+        {
+            return QSettings().value("UI/highlight_color", "orange");
+        }
     }
     return QVariant();
 }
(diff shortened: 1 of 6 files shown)


client/models/messageeventmodel.cpp
```diff
@@ -228,36 +236,48 @@ QVariant MessageEventModel::data(const QModelIndex& index, int role) const
...
+            // The below code assumes senderName output in AuthorRole
             switch( e->membership() )
             {
                 case MembershipType::Join:
-                    return QString("%1 (%2) joined the room").arg(e->displayName(), e->userId());
+                    return tr("joined the room");
                 case MembershipType::Leave:
-                    return QString("%1 (%2) left the room").arg(e->displayName(), e->userId());
+                    if (e->senderId() != e->userId())
+                        return tr("doesn't want %1 in the room anymore").arg(subjectName);
```
Fxrh: Nice euphemism :D

KitsuneRal: Glad you liked it :)

KitsuneRal:
I'll try to bring back orange highlights while still catering somewhat to dark scheme users. Do you still want highlights in the timeline in inverted colours or you're ok if it follows the same palette as in the room list?

Fxrh:
Orange on white is good, too :)

KitsuneRal:
There you go :) Orange on black looks terrific as well.

KitsuneRal:
...but the case of theme switching is now corrupted. Let me give one more commit to it.

KitsuneRal:
Turns out it's not a problem in the code, it's a Qt short coming. My quick guess is that `SystemPalette` QML class has no way of knowing about palette changes - QPalette is not a QObject and has no signals and slots. Given that without palette changes the thing looks as expected - please consider this ready for merging.

Fxrh:
Very nice, thanks! Yes, I've seen many applications that needs a restart after a system palette change.

PR #84, opened 2016-10-05
Don’t show timezone in chatwindow timestamps

maralorn:
Format is now. <hh:mm:ss>


KitsuneRal:
Sorry, that doesn't quite work with me. How about locales with AM/PM, e.g.?
I fully agree on the original point that timezone is completely foreign in that. @Fxrh, how much do we want to display seconds, actually? If we don't, Locale.ShortFormat will perfectly serve the purpose instead of hardwiring the pattern string.

Issue #80, opened 2016-10-04
Rooms with only leave/joins/nick changes/other unimportant stuff should be marked unread in a less agressiv manner, than with an unread message.


maralorn:

maralorn:
Closed by #87.

PR #87, opened 2016-10-06
Mark a room unread only if the new events are messages

maralorn:


maralorn:
Oh, btw. this does not work perfectly. When loading the program any room with a (real) message in the initialsync will be marked unread. Quaternion removes the unread flag, if it receives a receipt for the room by the current user. This has to be a receipt for the last event in the room, to remove the flag. If there were events but no messages since the last receipt, the room will stay unread, even if the user already read all (real) messages. Fixing this would require a bit more tinkering with the event chain.

Fxrh:
Currently, we have:
```
if( m_shown && m_unreadMessages )
     if( !messageEvents().empty() )
         markMessageAsRead( messageEvents().last() );
```

We'd need to change that to:
```
if( m_shown && !messageEvents().empty() && lastReadEvent(connection()->user()) != messageEvents().last() )
    markMessageAsRead( messageEvents().last() );
````
right?

maralorn:
Oh, yeah. Missed that one. (Apart from this still being a horrible way to set the read marker. Ask @mweinelt)

maralorn:
My latest variation on the commit should address all the previously mentioned concerns.

Was it a good to idea to do the whole change in one commit? It did this way because no change makes sense without the others.

maralorn:
This fixes #80.

KitsuneRal:
I'm basically fine (already looked at it, in fact) but I'd use iterators in the last loop (`for (auto it = messageEvents.rbegin(); it != messageEvents.rend(); ++it)`, saving on `at()` calls inside the loop). But here it's really a matter of preference, I think.

maralorn:
I’d like that. But I would have to read up on the iterators and I don’t have time for it right now.

maralorn:
So, I tried (succeeded) to implement that. See the other PR #95, but this only runs with QT 5.6. I think we should just merge it like that.

PR #89, opened 2016-10-09
Don‘t send everything the user enters, if it is probably unintended

maralorn:
Don’t send unknown commands and empty messages.

The display of the warning to the user is slightly buggy.
(Meaning the message might disappear too fast, if it is displayed multiple times in 5 seconds.)
But this is probably only a temporary solution anyways, until we get local echo.

On the other hand I am not sure if we really want warnings like this permanently in the timeline.
We could give it a timeout in the timeline, which would probably be pretty neat. We could also do typing notifications with that.

### DIFF
--- a/client/chatroomwidget.cpp
+++ b/client/chatroomwidget.cpp
@@ -217,6 +217,17 @@ void ChatRoomWidget::sendLine()
             {
                 text.remove(0, 3);
                 m_currentConnection->postMessage(m_currentRoom, "m.emote", text);
+            }
+            else if( text.startsWith("//") )
+            {
+                text.remove(0, 1);
+                m_currentConnection->postMessage(m_currentRoom, "m.text", text);
+            }
+            else if( text.startsWith("/") )
+            {
+                m_currentlyTyping->setText( "Unknown command. Use // to send this line literally" );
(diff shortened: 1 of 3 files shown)


client/chatroomwidget.cpp
```diff
@@ -187,6 +187,17 @@ void ChatRoomWidget::sendLine()
...
                 m_currentConnection->postMessage(m_currentRoom, "m.emote", text);
+            }
+            else if( text.startsWith("//") )
+            {
+                text.remove(0, 1);
+                m_currentConnection->postMessage(m_currentRoom, "m.text", text);
+            }
+            else if( text.startsWith("/") )
+            {
+                m_currentlyTyping->setText( "Unknown command. Use // to send this line literally" );
```
Fxrh: Hmm, maybe this would make more sense to put it into the status bar?

Fxrh: That would probably need some signal/slot mechanisms (having an showErrorMessage(QString) or showStatusMessage(QString) signal and connect that to the showMessage slot of the status bar)

maralorn: Aaaah, right. Wait a sec.

maralorn:
Wow, that is really neat to program. I like it.

Issue #46, opened 2016-08-31
Auto-complete names by pressing tab


PureTryOut:
Just like the Vector client. Basically currently pressing tab moves focus out of the chat line, but it should probably auto-complete the nickname you're typing instead.

KitsuneRal:
Yes, this is most-wanted but not easy to implement properly. But yeah, probably it's time to take a bite.

PureTryOut:
For some reason this wasn't closed automatically once #92 got merged. I guess the text "fixes #46" had to be in the initial PR message?

PR #92, opened 2016-10-10
Add tabcompletion

maralorn:
So this is tab completion, how I like it most.
To test it: Just hit Tab, whenever in the input bar.
I think it’s very usable this way.

I never wrote productive C++ code before.
So please criticize me for coding style, memory leaks or inefficiencies however you want.
I am open to change about everything in this PR.

### DIFF
--- a/client/chatroomwidget.cpp
+++ b/client/chatroomwidget.cpp
@@ -253,7 +253,7 @@ void ChatRoomWidget::triggerCompletion()
     if ( m_completing )
     {
         m_currentlyTyping->setText( QString("<i>Tab Completion (next: %1)</i>").arg(
-            m_completionList.mid( (m_completionListPosition + 1) % m_completionList.length(), 5).join(", ") ) );
+            QStringList(m_completionList.mid( (m_completionListPosition + 1) % m_completionList.length(), 5)).join(", ") ) );
         QString inputText = m_chatEdit->text();
         m_chatEdit->setText( inputText.left(m_completionInsertStart)
             + m_completionList.at(m_completionListPosition)
--- a/client/chatroomwidget.cpp
+++ b/client/chatroomwidget.cpp
@@ -252,8 +252,6 @@ void ChatRoomWidget::triggerCompletion()
     }
     if ( m_completing )
     {
-        m_currentlyTyping->setText( QString("<i>Tab Completion (next: %1)</i>").arg(
-            QStringList(m_completionList.mid( (m_completionListPosition + 1) % m_completionList.length(), 5)).join(", ") ) );
         QString inputText = m_chatEdit->text();
         m_chatEdit->setText( inputText.left(m_completionInsertStart)
             + m_completionList.at(m_completionListPosition)
@@ -261,6 +259,8 @@ void ChatRoomWidget::triggerCompletion()
         m_completionLength = m_completionList.at(m_completionListPosition).length();
         m_chatEdit->setCursorPosition( m_completionInsertStart + m_completionLength + m_completionCursorOffset );
         m_completionListPosition = (m_completionListPosition + 1) % m_completionList.length();
+        m_currentlyTyping->setText( QString("<i>Tab Completion (next: %1)</i>").arg(
+            QStringList(m_completionList.mid( m_completionListPosition, 5)).join(", ") ) );
     }
 }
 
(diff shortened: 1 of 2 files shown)


maralorn:
This fixes #46

client/chatroomwidget.cpp
```diff
@@ -190,3 +218,92 @@ void ChatRoomWidget::sendLine()
...
+        m_chatEdit->setText( inputText.left(m_completionInsertStart)
+            + m_completionList.at(m_completionListPosition)
+            + inputText.right(inputText.length() - m_completionInsertStart - m_completionLength) );
+        m_completionLength = m_completionList.at(m_completionListPosition).length();
+        m_chatEdit->setCursorPosition( m_completionInsertStart + m_completionLength + m_completionCursorOffset );
+        m_completionListPosition = (m_completionListPosition + 1) % m_completionList.length();
+    }
+}
+
+void ChatRoomWidget::startNewCompletion()
```
KitsuneRal: Generally here: it's better to use prefix in/decrement instead of postfix (the shortest explanation is at http://www.gotw.ca/gotw/002.htm, point 3).

client/chatroomwidget.cpp
```diff
@@ -190,3 +218,92 @@ void ChatRoomWidget::sendLine()
...
+{
+    QString inputText = m_chatEdit->text();
+    int cursorPosition = m_chatEdit->cursorPosition();
+    for ( m_completionInsertStart = cursorPosition - 1;
+          m_completionInsertStart >= 0;
+          m_completionInsertStart-- )
```
KitsuneRal: This for() can be rewritten in two lines:
`for( m_completionInsertStart = cursorPosition;
 --m_completionInsertStart >= 0; )`

client/chatroomwidget.cpp
```diff
@@ -190,3 +218,92 @@ void ChatRoomWidget::sendLine()
...
+    findCompletionMatches(inputText.mid(m_completionInsertStart, m_completionLength));
+    if ( !m_completionList.isEmpty() )
+    {
+        m_completionCursorOffset = 0;
+        m_completionListPosition = 0;
+        m_completing = true;
+        m_completionLength = 0;
+        if ( m_completionInsertStart == 0)
+        {
+            m_chatEdit->setText(inputText.left(m_completionInsertStart) + ": " + inputText.right(inputText.length() - cursorPosition));
```
KitsuneRal: You can use `QString::mid` instead of `QString::right` here and further below, avoiding calculation of the number of characters to pick.

client/chatroomwidget.cpp
```diff
@@ -190,3 +218,92 @@ void ChatRoomWidget::sendLine()
...
+void ChatRoomWidget::triggerCompletion()
+{
+    if ( !m_completing && m_currentConnection && m_currentRoom )
+    {
+        startNewCompletion();
+    }
+    if ( m_completing )
+    {
+        m_currentlyTyping->setText( QString("<i>Tab Completion (next: %1)</i>").arg(
+            QStringList(m_completionList.mid( (m_completionListPosition + 1) % m_completionList.length(), 5)).join(", ") ) );
```
KitsuneRal: It seems to me that you have exactly the same expression in the end of this block. Why not reuse it, moving `m_currentlyTyping->setText()` to the very end, after updating `m_completionListPosition`?

client/chatroomwidget.cpp
```diff
@@ -252,30 +252,28 @@ void ChatRoomWidget::triggerCompletion()
...
 }
 
 void ChatRoomWidget::startNewCompletion()
 {
     QString inputText = m_chatEdit->text();
     int cursorPosition = m_chatEdit->cursorPosition();
-    for ( m_completionInsertStart = cursorPosition - 1;
-          m_completionInsertStart >= 0;
-          m_completionInsertStart-- )
+    for ( m_completionInsertStart = cursorPosition - 1; --m_completionInsertStart >= 0; )
```
KitsuneRal: It should be `cursorPosition` instead of `cursorPosition - 1`, I suppose. `--m_completionInsertStart >= 0` will be calculated (and the decrement will occur therefore) before even the first iteration.

maralorn: Hmpf, should have seen that …

maralorn:
If you wish I can clean up and force push this.

KitsuneRal:
No need, if you ask me. 6 commits don't take too much space in the tree. @Fxrh?

Fxrh:
Imho the commits don't need to be rebased.

Issue #67, opened 2016-09-19
Only mark messages as read after user interaction


Fxrh:
Currently, Quaternion marks all the incoming messages as read in the selected room. We have no idea, though, whether the user has read those messages (i.e., the user could have left his PC running while being away). Some action (like a mouse move event) should be required to mark messages as read.

maralorn:
Closed by #95.

PR #95, opened 2016-10-10
Detect activity to not set readmarker too fast

maralorn:
If you don’t like this, regard it as a POC.
The architecture of this changes should allow easy extension to set the read marker to the end of the currently shown timeline as opposed to the end of the complete timeline as it is now.

There are concerns in the Docs, that this kind of EventFilter slows the program down.
But I don’t see another solution. And it still feels snappy to me.^^

(Note: This extends my other PR regarding marking rooms unread.)

### DIFF
--- a/client/chatroomwidget.cpp
+++ b/client/chatroomwidget.cpp
@@ -115,6 +115,12 @@ ChatRoomWidget::~ChatRoomWidget()
 {
 }
 
+void ChatRoomWidget::lookAtRoom()
+{
+    if ( m_currentRoom )
+        m_currentRoom->lookAt();
+}
+
 void ChatRoomWidget::enableDebug()
 {
     QQmlContext* ctxt = m_quickView->rootContext();
--- a/client/chatroomwidget.h
+++ b/client/chatroomwidget.h
@@ -47,6 +47,7 @@ class ChatRoomWidget: public QWidget
         void enableDebug();
         void triggerCompletion();
         void cancelCompletion();
+        void lookAtRoom();
 
     signals:
         void joinRoomNeedsInteraction();
--- a/client/main.cpp
+++ b/client/main.cpp
@@ -37,7 +37,7 @@ class ActivityDetector : public QObject
             switch (ev->type())
             {
             case QEvent::KeyPress:
-            case QEvent::ApplicationActivate:
+            case QEvent::FocusIn:
             case QEvent::MouseMove:
             case QEvent::MouseButtonPress:
                 m_mainWindow->activity();
--- a/client/main.cpp
+++ b/client/main.cpp
@@ -24,6 +24,32 @@
 
 #include "mainwindow.h"
 
+class ActivityDetector : public QObject
+{
+    public:
+        ActivityDetector(MainWindow* c): m_mainWindow(c)
+        {
+            c->setMouseTracking(true);
+        };
+    protected:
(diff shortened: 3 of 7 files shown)


maralorn:
This fixes #67 btw.

KitsuneRal:
Well, I have several issues here:
- Why not just override `QMainWindow::event()`, and call `setMouseTracking` directly from `main()`?
- `QEvent::ApplicationActivate` is deprecated, as of Qt 5.7. Probably, `QEvent::FocusIn` on `QMainWindow` would be a better substitute.
- Did I get it correct that in order the next message to be marked as read I should do something actively (move a mouse, in particular)? That doesn't seem right. I don't necessarily use input devices when I read messages. At least, I'd prefer this to be check after a couple of minutes of idleness, with messages auto-marked read before that.
- One more idea (I don't know if it works or not) is to override `ChatRoomWidget::paint()`, assuming that `paint()` is only called when a widget is exposed and/or its contents change. This more precisely catches the moment of showing a particular message to the user. If the user looks the other way at that moment, we can't help it; but we can't track movement of the user's eyes without being creepy, anyway.

maralorn:
1. I tried the `event()` function of the main window first. It only get’s events which are not consumed by any contained widget. That is definitely not enough.
2. Ooops, I hate it, when search machines give me outdated documentation. I would perhaps be more bold and use `QEvent::ApplicationStateChange`?
3. Yes, you got that right. I think this is completely reasonable behaviour. (And it is what Riot does as far as I can tell.) You sitting in front of the computer doing absolutely nothing with it for significant time while reading one channel seems unlikely to me. At least it is very probable, that you are afk. I think marking it as read, after a few minutes is definitely a bad idea. Since this will get triggered, whenever you are afk. If you want to mark read in the complete absences of user interaction I would do it nearly instantly. Then the criterion should be to discover as good as possible, that the message is really shown on a display.
4. This sounds like a hack. Isn’t there a signal for that or something? But if this actually is closer to the expected behaviour I would be up to it.

Fxrh:
w.r.t. 3: I don't think that a timeout is a valid option, as if I leave my computer, it would still mark some messages as read and then stop at some point, which would give a false impression on what the last message was.

While we will never know for sure whether the user has read a message or not, I think both checking for user activity and checking for the window being shown are good heuristics. So now we have to decide whether we prefer false positives or false negatives... As we are using the last read message to decide whether we display a room to contain new messages, I'm leaning a bit more to not mark messages as read in ambiguous cases.

w.r.t. 4: I'm not sure this works with modern window managers providing transparency, preview images of windows and such things..

maralorn:
w.r.t 3: I want to add one argument here. The implemented method is eventually consistent. There is no way for the user (killing the program on purpose aside) to switch the room away without marking it read. So this method will give no false positives and the false negatives will only be very temporary.

maralorn:
Nice! But there is one corner case I discovered. (I am testing this patch in my "productive" chatting.) We should mark our own message always as read. This does not happen at the moment, if you don’t move after hitting enter.

maralorn:
Okay, imo, this is ready for merge.

KitsuneRal:
Sorry for the silence, got carried away with other things a bit. Actually, I think we can simply auto-read everything with `event->sender() == connection->user()->id()`. Local echo is not necessary for that.

PR #96, opened 2016-10-10
Make the scrolldown indicator translucent

KitsuneRal:


maralorn:
If I may comment on that: I think it is a good idea.

KitsuneRal:
Yeah, already opened a separate PR from myself on that (which was clumsy a undoing a merge would be even clumsier).

PR #101, opened 2016-10-19
Followup on the recent merges

KitsuneRal:
1. I used iterators in that code in `QuaternionRoom`. Since there's no easy way to deal with reverse iterators in older Qt, I used ordinary ones; since I formally can't decrement before begin(), I used an easy workaround, decrementing the iterator the first thing in the loop body instead of `for` statement.
2. It turned out that ListView.atYEnd in older Qts (at least 5.2.1) ceases to update after one true/false/true turnaround. So I reimplemented the property - the home-grown one works flawlessly (although, probably, it's a little harder for the CPU).

### DIFF
--- a/client/quaternionroom.cpp
+++ b/client/quaternionroom.cpp
@@ -120,16 +120,17 @@ void QuaternionRoom::processEphemeralEvent(QMatrixClient::Event* event)
     if ( m_unreadMessages && event->type() == QMatrixClient::EventType::Receipt )
     {
         QString lastReadId = lastReadEvent(connection()->user());
-        for (int i = messageEvents().size()-1; i >= 0; i--)
+        for (auto it = messageEvents().end(); it != messageEvents().begin(); )
         {
-            if ( lastReadId == messageEvents().at(i)->id() )
+            --it;
+            if ( lastReadId == (*it)->id() )
             {
                 m_unreadMessages = false;
                 emit unreadMessagesChanged(this);
                 qDebug() << displayName() << "no unread messages";
                 break;
             }
-            if ( messageEvents().at(i)->type() == QMatrixClient::EventType::RoomMessage )
+            if ( (*it)->type() == QMatrixClient::EventType::RoomMessage )
                 break;
         }
     }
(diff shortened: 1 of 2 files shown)


client/quaternionroom.cpp
```diff
@@ -88,14 +88,17 @@ void QuaternionRoom::doAddNewMessageEvents(const QMatrixClient::Events& events)
...
     {
         m_messages.push_back(makeMessage(e));
         if ( e->type() == QMatrixClient::EventType::RoomMessage )
             new_message = true;
         if ( e->senderId() == connection()->userId() )
-            markMessageAsRead( e );
+            lastOwnMessage = e;
     }
+    if (lastOwnMessage)
+        markMessageAsRead( lastOwnMessage );
```
maralorn: I am confused. So we need this, although we should not send our own readmarker to the server, because the server won’t return us a readmarker? So marking our own messages as read … Shouldn’t that be something the library does?

Will calling `markMessagAsRead` trigger `processEphemeralEvent` of this class? If it does it instantly, this can cause a bug, because the room will be marked read here, because our own message is the last, but will be marked unread again in the next line of this function.

What do you think?

maralorn: I just tested: That bug does happen. The room will be in the state, that the last line is marked read, but the room will show as unread in the list, reproducible by sending a message with another client while looking at the room.

maralorn: The fastest fix would be to move this two lines to the end of the method.

KitsuneRal: Right you are :-/ Those lines should be in the end of the method. Also there's one more problem - if a sync has other's messages followed by an own message, the last one is marked as auto-read but a read receipt for previous ones (those from others) is not sent. This will be fixed in the library code.

KitsuneRal: As I said, it's actually a problem in the library code (see the kitsune-read-receipts branch): as of now I just don't send any read receipt if the message to mark read belongs to the local user. Instead it should mark as read the local user's message _and_ send a read receipt for the last other's message before the local user's one(s).

maralorn: Aaaaah, okay. Yeah, that is a library bug. Thanks for explaining.

PR #102, opened 2016-10-20
A quick and dirty implementation of a read marker line

KitsuneRal:
It happened partially as a by-product of read receipts digging - I realized I want to see which event is currently marked as read. Accidentally it turned out to be quite usable from the onset :)
Note: this relies on kitsune-fixes: first three commits are from there.

### DIFF
--- a/client/quaternionroom.cpp
+++ b/client/quaternionroom.cpp
@@ -120,16 +120,17 @@ void QuaternionRoom::processEphemeralEvent(QMatrixClient::Event* event)
     if ( m_unreadMessages && event->type() == QMatrixClient::EventType::Receipt )
     {
         QString lastReadId = lastReadEvent(connection()->user());
-        for (int i = messageEvents().size()-1; i >= 0; i--)
+        for (auto it = messageEvents().end(); it != messageEvents().begin(); )
         {
-            if ( lastReadId == messageEvents().at(i)->id() )
+            --it;
+            if ( lastReadId == (*it)->id() )
             {
                 m_unreadMessages = false;
                 emit unreadMessagesChanged(this);
                 qDebug() << displayName() << "no unread messages";
                 break;
             }
-            if ( messageEvents().at(i)->type() == QMatrixClient::EventType::RoomMessage )
+            if ( (*it)->type() == QMatrixClient::EventType::RoomMessage )
                 break;
         }
     }
--- a/client/quaternionroom.cpp
+++ b/client/quaternionroom.cpp
@@ -97,7 +97,7 @@ void QuaternionRoom::doAddNewMessageEvents(const QMatrixClient::Events& events)
             new_message = true;
     }
     if (lastOwnMessage)
-        setLastReadEvent(connection()->user(), lastOwnMessage->id());
+        promoteReadMarker(connection()->user(), lastOwnMessage->id());
 
     if( !m_unreadMessages && new_message)
     {
--- a/client/quaternionroom.cpp
+++ b/client/quaternionroom.cpp
@@ -88,14 +88,17 @@ void QuaternionRoom::doAddNewMessageEvents(const QMatrixClient::Events& events)
 
     m_messages.reserve(m_messages.size() + events.size());
     bool new_message = false;
+    QMatrixClient::Event* lastOwnMessage = nullptr;
     for (auto e: events)
     {
         m_messages.push_back(makeMessage(e));
         if ( e->type() == QMatrixClient::EventType::RoomMessage )
             new_message = true;
         if ( e->senderId() == connection()->userId() )
-            markMessageAsRead( e );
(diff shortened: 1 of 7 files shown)


maralorn:
Nice! Once it’s merged I will try to supply a nice animation for that.

KitsuneRal:
Just for the record: the failing VS2013 build is fixed in libqmatrixclient master but not merged into `kitsune-read-receipts` branch that this PR refers to. Once the libqmatrixclient PR is merged, I'll add a trivial commit here to refer to the correct libqmatrixclient commit.

client/quaternionroom.cpp
```diff
@@ -88,17 +87,18 @@ void QuaternionRoom::doAddNewMessageEvents(const QMatrixClient::Events& events)
...
     m_messages.reserve(m_messages.size() + events.size());
     bool new_message = false;
-    QMatrixClient::Event* lastOwnMessage = nullptr;
     for (auto e: events)
     {
         m_messages.push_back(makeMessage(e));
+        if (e->senderId() == connection()->userId())
+        {
+            if (!new_message)
+                setLastReadEvent(connection()->user(), e->id());
```
maralorn: I’m sorry. Why the `(!new_message)`? I don’t understand it at all. And I feel fairly certain about it causing bugs. But the negation would also not work.
If we are okay, with accessing the hashmap on every own new message, we can do it like this. But the condition shouldn’t be there in my opinion.

maralorn: I just tested. This leads to all rooms being marked as unread after the initial sync, in which I send the last message. There are 2 reasons for this:
1. The readmarker is not set correctly because there are normally new messages before our own.
2. We don’t call anything that would trigger `processEphemeralEvent` after this message loading, which was until now responsible for marking the room read.

KitsuneRal: This turns to be a hard one...
The idea of that double `if` is to automatically mark read (but only locally, not on the server) the user's own messages - and only until the first non-local message.
`processEphemeralEvent` is supposed to process events that have come from the server - in particular, read receipts; in particular, read receipts of the local user that originated from other devices. If marking is invoked locally (as in the case), it should not be used.
I reproduced the issue: a room gets unread after a restart if the last message in it is mine. The fix is in the works.

maralorn: We would we only mark own messages as read until the first non-local message? I think the semantics of the readmarker suggest, that we definitely „have read“ all messages before any of our own. The problem here is, that on load all rooms get marked unread because of all messages and then we selectively mark rooms read because the readmarkers from the initial sync get processed after the message events. But in more or less every initial sync there will be a non-local message. That means the code path here is very rarely followed.

maralorn: I fear that this use case is not supported by the matrix spec. Besides atm quaternion will mark the whole room read in the second you throw a glance at it. So there would be a lot more changes necessary to make that work. (I’d say it can not be done reasonably.)

KitsuneRal: Seems so, the more I think about it...

KitsuneRal:
Ok, next attempt to fix things. Seems good on previous test-cases (I didn't test it quite thoroughly, I admit).

maralorn:
Sooo … We might have stumbled over an old bug here, which we only really discover now, because the readline marker is new. Your fix is a step in the right direction, but it is still buggy.
Rooms in which I wrote the last message still are marked unread and the my own messages at the end don’t have the readmarkers set on load.
I have the following theory: The above function correctly sets the readmarker to the end of the timeline. But after that the ephemeral events from the server get processed which set the readmarker on the last non-local message, which is earlier in the timeline, therefore spoiling our effort. I am not sure how to debug that. Can anyone confirm, that the server behaves like that? Or does the library prevent the readmarker from going backwards? (which feels hard to do)

KitsuneRal:
Actually, it's not very hard to prevent the readmarker to retract. And, I suppose, it's a good thing for now in any case. I think I'll go that way.

KitsuneRal:
@maralorn, does the last commit work for you?

maralorn:
This is by far the best solution. Very nice. I am not 100% convinced that it is semantically perfect, but I cannot find any misbehaviour.

On thing: I think the internal advancing of the readreceipt for own messages should a) be done in the library and b) for every user, automatically whenever we get new events.

But please merge this!

KitsuneRal:
Actually, I even think that the whole unread messages stuff should go to the library. The current split is rather unnatural. I just wanted to debug the damn read markers thing first.

KitsuneRal:
Thanks to @maralorn, the read marker is now animated; I also switched its colour to "highlight" to make it a bit more prominent. And AppVeyor is finally happy because I updated to the newer lib commit that builds with VS2013. Ready to merge.

PR #107, opened 2016-10-23
Fix scrollToBottom

### LINKED ISSUE #61: Timeline with images doesn't keep to the bottom
When a timeline with images is opened, images take time to load. When they actually load and expand in the timeline, the part below an image is displaced further down, which is not an expected behaviour at least when a user is standing on the newest (bottommost message).

This is one of consequences of having a beginning of the model on the top, rather than on the bottom. Unfortunately, reversing the model and the chat ListView (with `verticalLayout` property) corrupts positioning of section labels (the labels appear below instead of above). Asking the Qt mailling list for help didn't help; seems that the only reasonable option left is to replace section labels with a home-baked solution for grouping by date (somewhat similar to the hack that I've made for the scrollbar).

maralorn:
1. rename wasAtEndY to stickToBottom because seriously AtEndY vs.
AtYEnd was driving me crazy.

2. Don’t use aboutToBeInserted this is unreliable. Infact this event
will sometimes be triggerted in an inconvenient situation of the
population and unstick the scroll erroneously. Instead we track all the
time if we want to stickToBottom. The default is yes. And it only
changes when the user scrolls manually (onMovementStarted/onMovementEnded)
FIXME: This state should be tracked per room and perhaps depend on
the readmarker

3. Use a Timer Component to trigger the scrollingdown whenever
something changes. We need to give the controlflow away and scroll
after QML has updated and loaded the view. This has to be iterated,
because the scroll may trigger another update and loading …

This fixes #61.

### DIFF
--- a/client/qml/chat.qml
+++ b/client/qml/chat.qml
@@ -12,8 +12,23 @@ Rectangle {
 
     signal getPreviousContent()
 
+    Timer{
+        id: scrollTimer
+        interval: 1;
+        onTriggered: reallyScrollToBottom()
+    }
+
+    function reallyScrollToBottom() {
+        if (chatView.stickToBottom && !chatView.nowAtYEnd)
+        {
+            chatView.positionViewAtEnd()
+            scrollToBottom()
+        }
+    }
+
     function scrollToBottom() {
-        chatView.positionViewAtEnd();
+        chatView.stickToBottom = true
+        scrollTimer.running = true
     }
 
     ListView {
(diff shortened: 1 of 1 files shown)


KitsuneRal:
I wonder if 1 ms timeout is even needed. Is 0 acceptable? In Qt Widgets, it's a known pattern to give up control to the event loop.

maralorn:
Oh, yeah, that should work.

maralorn:
Oh, the heck … Don’t merge this yet, I am experiencing serious performance issues.

maralorn:
Okay, I don’t know what the heck was going on. The default timer interval does not seem to be the one from the docs. Not I works and actually does not feel glitchy anymore.

PR #109, opened 2016-10-23
Two very simple fixes

KitsuneRal:
Don't reset the room if it's the same as before; and a QML warning fix


KitsuneRal:
This will conflict with the contents of #102; will rebase and re-issue after merging that one.

PR #112, opened 2016-10-24
Detect activity

### LINKED ISSUE #104: CPU usage jumps while mouse is moving over window
After the room list has loaded, clicking on a room and moving the mouse around the window causes `Job "PostReceiptJob" created` to be output rapidly which also causes the cpu usage of Quaternion to jump.

Tested on Ubuntu 16.10

maralorn:
WARNING: This PR includes code from the recent readmarker branch. (which is ready to be merged afaik)

So now I made the ActivityDetector more fancy, it will only be enabled when the currently shown room is unread. I think now the impact on performance is neglectable.

Fixes #104

### DIFF
--- a/client/activitydetector.cpp
+++ b/client/activitydetector.cpp
@@ -1,0 +1,67 @@
+/**************************************************************************
+ *                                                                        *
+ * Copyright (C) 2016 Malte Brandy <malte.brandy@maralorn.de>                        *
+ *                                                                        *
+ * This program is free software; you can redistribute it and/or          *
+ * modify it under the terms of the GNU General Public License            *
+ * as published by the Free Software Foundation; either version 3         *
+ * of the License, or (at your option) any later version.                 *
+ *                                                                        *
+ * This program is distributed in the hope that it will be useful,        *
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of         *
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the          *
+ * GNU General Public License for more details.                           *
+ *                                                                        *
+ * You should have received a copy of the GNU General Public License      *
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.  *
+ *                                                                        *
+ **************************************************************************/
+
+#include "activitydetector.h"
+#include "mainwindow.h"
+#include "chatroomwidget.h"
+#include "models/messageeventmodel.h"
+#include <QtCore/QDebug>
+
+ActivityDetector::ActivityDetector(QApplication* a, MainWindow* w): m_app(a), m_mainWindow(w), m_enabled(false), m_messageEventModel(w->getChatRoomWidget()->getMessageEventModel())
+{
+    connect(this, &ActivityDetector::triggered, m_messageEventModel, &MessageEventModel::markShownAsRead);
+    connect(m_messageEventModel, &MessageEventModel::lastShownIndexChanged, this, &ActivityDetector::updateEnabled);
+    connect(m_messageEventModel, &MessageEventModel::readMarkerIndexChanged, this, &ActivityDetector::updateEnabled);
+}
+
+void ActivityDetector::updateEnabled()
+{
+    setEnabled(m_messageEventModel->awaitingMarkRead());
+}
+
+void ActivityDetector::setEnabled(bool enabled)
+{
+    if (enabled && !m_enabled) {
+        m_app->installEventFilter(this);
+        m_mainWindow->setMouseTracking(true);
+        m_enabled = true;
+        qDebug() << "enabling ActivityDetector";
+    }
+    if (!enabled && m_enabled) {
+        m_mainWindow->setMouseTracking(false);
+        m_app->removeEventFilter(this);
+        m_enabled = false;
+        qDebug() << "disabling ActivityDetector";
+    }
+}
+
+bool ActivityDetector::eventFilter(QObject* obj, QEvent* ev)
+{
+    switch (ev->type())
+    {
+    case QEvent::KeyPress:
+    case QEvent::FocusIn:
+    case QEvent::MouseMove:
+    case QEvent::MouseButtonPress:
+        emit triggered();
+        setEnabled(false);
+    default:;
+    }
+    return QObject::eventFilter(obj, ev);
+}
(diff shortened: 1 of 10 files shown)


maralorn:
I fixed the PR as discussed above.

KitsuneRal:
One cosmetic thing - how about using `#pragma once` instead of those 20th-century `#ifdef` header file guards? It works everywhere for some 10 years already.

maralorn:
Uh, oh … I just copied that from all the other files … Wait a sec.

KitsuneRal:
In fact, I'd gladly change it in all other files as well but still wait for I don't know what :)

maralorn:
Well it seems tedious …

client/activitydetector.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include "mainwindow.h"
+#include "chatroomwidget.h"
+#include "models/messageeventmodel.h"
+#include <QtCore/QDebug>
+
+ActivityDetector::ActivityDetector(QApplication* a, MainWindow* w): m_app(a), m_mainWindow(w), m_enabled(false), m_messageEventModel(w->getChatRoomWidget()->getMessageEventModel())
+{
+    connect(this, &ActivityDetector::triggered, m_messageEventModel, &MessageEventModel::markShownAsRead);
+    connect(m_messageEventModel, &MessageEventModel::lastShownIndexChanged, this, &ActivityDetector::updateEnabled);
+    connect(m_messageEventModel, &MessageEventModel::readMarkerIndexChanged, this, &ActivityDetector::updateEnabled);
```
KitsuneRal: This hints that read marker and last shown index are two things that should in fact be one thing. That's not a comment on this code, rather on my implementation in chat.qml and `MessageEventModel`. As talked in #quaternion, this should be a single property.

client/activitydetector.h
```diff
@@ -0,0 +1,49 @@
...
+        void triggered();
+
+    protected:
+        bool eventFilter(QObject* obj, QEvent* ev);
+
+    private:
+        QApplication* m_app;
+        MainWindow* m_mainWindow;
+        bool m_enabled;
+        MessageEventModel* m_messageEventModel;
```
KitsuneRal: If you use lambdas in the `connect`s, you can get rid of both this member variable and `updateEnabled()` method. Reduces coupling with the chat widget innards.

KitsuneRal:
Before working on this, let me unify two properties into one.

client/activitydetector.cpp
```diff
@@ -0,0 +1,67 @@
...
+ *                                                                        *
+ **************************************************************************/
+
+#include "activitydetector.h"
+#include "mainwindow.h"
+#include "chatroomwidget.h"
+#include "models/messageeventmodel.h"
+#include <QtCore/QDebug>
+
+ActivityDetector::ActivityDetector(QApplication* a, MainWindow* w): m_app(a), m_mainWindow(w), m_enabled(false), m_messageEventModel(w->getChatRoomWidget()->getMessageEventModel())
```
maralorn: I am a bit confused. I think you or @fxrh told me some days ago the logic shouldn’t be in the chatroom widget. Because it is just a collection of widgets or something …
But I actually agree …

client/activitydetector.cpp
maralorn: I think having the readMarkerIndex is very handy in the QML. I use it for scroll indicator business in another PR.

maralorn:
Okay, I will hold of. Mind my suggestion to use some properties like highestShownIndex and lowestShownIndex, which probably have to be set in QML but can the be used very neatly in C++. To achive exactly what we want. But I don’t mind if you find a better solution. (I think switching to indices is quite smart because it solves the ordering problems.)

client/activitydetector.cpp
KitsuneRal: Yes, that was me. In fact, my current implementation doesn't quite support my idea from then, and my feedback here only deals with my current+your proposed code. Sorry for confusing.

Speaking of ideals, what I'd like to ultimately see is a completely un-knowledgeable `ActivityDetector` object, that only tracks activity and fires a signal when needed; and the whole orchestration (i.e. connecting signals) localised in some higher level class, preferably `MainWindow`. But in order to do that, we at least need to expose all necessary information in `Room` (or, rather, `QuaternionRoom`, as it heavily relies on Quaternion's way of showing the timeline), not in `MessageEventModel`.

client/activitydetector.cpp
KitsuneRal: Enumerating events would be very handy in general... I already think of having events that are aware of their own whereabouts inside the timeline. But that's a separate topic. Fully understand that to scroll to read marker we need an index in QML.

KitsuneRal:
I looked through your commit again, and I've got one more question: will it work if you use comparison between `Room::lastReadEvent()` and `Room::messageEvents()[lastShownIndex]->id()` in the meantime instead of `readMarkerIndex < lastShownIndex`?
I'm actually planning to make events navigable and order-comparable in the library - I think I've found a fairly decent and compact way to do it.

maralorn:
If you implement a correct compare operator for events, that should totally work.

So how do I procede now? You fixed the issue we talked about, but you didn’t actually unify the 2 properties. Which changes do you want me to do now?

maralorn:
I can move the logic where ever you want. I’d suggest the main window. But I will have to insert something in some header files for that, because the activity detector is not know to the rest of the program.

KitsuneRal:
Regarding comparison - actually, what I meant to say is to compare event id's for equality, rather than events for their order  But it turned out that navigating and comparing events in the timeline is not so difficult, though it required switching to `std::deque`, as I mentioned in #quaternion.

I'd really prefer this to get in together with my better-read-marker branch - can you make another PR to my repo so that we could then make one consolidated PR out of our two? I'm rather cautious to go with `std::deque` yet; so probably let's try to go conservative and simply merge the two PRs into one. As for the placement of `ActivityDetector` connections - I'd also like to see that setup in `MainWindow`. If you need something from the lib for that, let's make it happen.

maralorn:
So I can make this PR against your repo, no problem. If I shall put the logic into the main window and fix the other stuff from above, that will have to wait till the weekend.

KitsuneRal:
Following up on that old topic - the events are navigable now (even in master libqmatrixclient). You will most likely miss an accessor or two in `Room` for this matter - feel free to bug/PR me as necessary.

KitsuneRal:
Superseded by #135

PR #114, opened 2016-10-27
More fixes

KitsuneRal:
Not having enough time for serious hacking, I took an opportunity to fix minor things here and there. Among highlights - `make install`, or rather, `cmake --build . --target install`, on Windows carry over the nevessary Qt libraries to the installation folder as well, so the whole build-install-run process becomes (almost) perfectly automated.

PR #115, opened 2016-10-30
Fix build instructions in README

Takios:
The last part should be "cmake --build ." because we already
cd'd into the build directory

This has been irking me since I've built Quaternion the first time...


KitsuneRal:
Thanks a lot but this has just been addressed in #114. Just wait for a merge.

PR #116, opened 2016-10-30
Don’t raise the mainwindow on highlight, just send an activateEvent

### LINKED ISSUE #113: Quaternion raises itself on notification
When the user gets highlighted (notified, whatever), Quaternion raises itself above all other windows and steals focus from whatever the user was doing. This is quite annoying, and a regular notification is the only thing that should be shown.

Happens on both Xubuntu and Arch (so OS independent really) using Xfwm (from the Xfce4 DE).

maralorn:
This should fix #113. But this needs testing with different WMs, to see if it is the intended behaviour.

Works like intended with i3wm.
I’d like to here from xfce, kde and windows.


PureTryOut:
Confirmed working on Xfce4 (Xfwm).

KitsuneRal:
FWIW: Works fine on Windows 7.

maralorn:
So if @fxrh is satisfied with the KDE behaviour this should be ready to merge.

PR #117, opened 2016-11-04
Better read marker

### LINKED ISSUE #110: Only mark events read that are currently in view.

KitsuneRal:
This is the more accurate read marker implementation. Specifically, it tracks which messages in the room can actually be seen at the moment and marks only those messages as read. Moreover, it only starts marking messages as read from the moment when the read marker appears on the screen. So the overall logic is now as follows:

- If the most recent read message is off the screen, no messages are marked as _shown_ or _read_.
- When the read marker is on the screen, messages until it are immediately marked as _shown_. Messages that are on the screen but below the read marker are marked as _shown_ after 1 second of staying on the screen.
- When the activity detector detects user interaction, all _shown_ messages are marked as _read_.
- There's a special behaviour for messages from the local user: they're automatically marked as _read_ but only if they're among already _read_ others' messages or if they _immediately_ follow the most recent message not from a local user.

Besides, probably, some fine-tuning of `ActivityDetector`, I consider this implementation production-ready. @maralorn, your input on `ActivityDetector` as well as on this PR, is much welcome.

Warning: this PR needs a change in libqmatrixclient that is not in `master` yet: Fxrh/libqmatrixclient#48. Once the libqmatrixclient PR is approved, the current PR will be updated to refer to `master`.

Fixes #110.

### DIFF
--- a/client/qml/chat.qml
+++ b/client/qml/chat.qml
@@ -138,49 +138,8 @@ Rectangle {
                 y + message.height - 1 > chatView.contentY &&
                 y + message.height - 1 < chatView.contentY + chatView.height
 
-            property bool newlyShown: shown &&
-                                      messageModel.lastShownIndex !== -1 &&
-                                      index > messageModel.lastShownIndex
-
-            function promoteLastShownIndex() {
-                if (index > messageModel.lastShownIndex) {
-                    // This doesn't promote the read marker, only the shown
-                    // index in the model. The read marker is updated upon
-                    // activity of the user, we have no control over that here.
-                    messageModel.lastShownIndex = index
-                    console.log("Updated last shown index to #" + index,
-                                "event id", eventId)
-                }
-            }
-
-            Timer {
-                id: indexPromotionTimer
-                interval: 1000
-                onTriggered: { if (parent.shown) promoteLastShownIndex() }
-            }
-
-            onShownChanged: {
-                if (messageModel.room.readMarkerEventId === eventId)
-                {
-                    if (shown)
-                        promoteLastShownIndex()
-                    else
-                        messageModel.lastShownIndex = -1
-                }
-            }
-
-            onNewlyShownChanged: {
-                // Only promote the shown index if it's been initialised from
-                // the read marker beforehand, and only if the message has been
-                // on screen for some time.
-                if (newlyShown)
-                {
-                    indexPromotionTimer.start()
-                    console.log("Scheduled lastReadIndex update from",
-                                messageModel.lastShownIndex, "to", index,
-                                "in", indexPromotionTimer.interval, "ms")
-                }
-            }
+            onShownChanged:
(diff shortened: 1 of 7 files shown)


maralorn:
This is awesome!
It is everything I dreamt for in terms of readmarker management.

I am not 100% convinced, that there really is no bug in here. I had one room erroneously marked unread once. But I can’t find a way to reproduce it.

Are you sure, that this does not lead to us loosing the read marker? What happens if our read marker is not inside the currently loaded messages?

I have a tiny bit of clean up and a lot of ideas now. I hope I’ll get to make the corresponding PRs.

maralorn:
I noticed one issue, which is not a regression, but still worth thinking about.
If you leave your computer while being scrolled to bottom, messages may appear, get marked shown and then get scrolled out of the window. Once you return, they will get marked as read, even though you probably never did.
I am trying to wrap my head around it, but I didn’t find an easy fix for that yet.

maralorn:
This fixes #110. \o/

KitsuneRal:
Uhmm, yeah, it's a trap. Either we have to intertwine the code that promotes the shown index with `ActivityDetector`, or we have to invalidate the shown index (set it back to -1) as soon as the read marker falls out of view.

KitsuneRal:
Thinking of it further - probably we could get rid of the "shown" concept completely if we engage `ActivityDetector` to determine whether the shown/read indicator should move or not.

KitsuneRal:
Ok, here's a quick fix for the read marker escaping out of view. I'm still unsure if "shown" and "read" shouldn't come together.

client/qml/chat.qml
```diff
@@ -132,6 +132,56 @@ Rectangle {
...
+            // viewing area of the timeline.
+            property bool shown:
+                y + message.height - 1 > chatView.contentY &&
+                y + message.height - 1 < chatView.contentY + chatView.height
+
+            property bool newlyShown: shown &&
+                                      messageModel.lastShownIndex !== -1 &&
+                                      index > messageModel.lastShownIndex
+
+            function promoteLastShownIndex() {
```
Fxrh: That's a lot of logic in QML... Currently, we use the QML part as UI, but defining which message should be set as read is not part of the UI, so wouldn't it make more sense to move that part to C++?

KitsuneRal: Well, we have to enable some kind of feedback from the QML view back to the C++ part anyway. But I can think how to reduce the amount of logic on the QML side. I guess you don't mind `shown` but are rather concerned about `newlyShown` and the things around it (such as `indexPromotionTimer`), correct?

Fxrh: yes

KitsuneRal:
As requested, I rewrote the newlyShown logic from QML to C++. It became a bit more verbose but assumingly more efficient.

PR #120, opened 2016-11-05
Improved scrollindicators

maralorn:
This should only be merged after @KitsuneRal ’s readmarker handling and my activity-detector change.

For the rest see commit message


maralorn:
Please feel free to comment on the design. I don’t think it’s perfect.

maralorn:
I assume this PR has bitrotten enough to just close it?

It's great to see, that Quaternion is still going strong, but obviously my focus has shifted.

KitsuneRal:
Nope, still has enough useful code so I keep it around. Thanks for contributing back then :)

PR #122, opened 2016-11-06
Remove quaternion message

maralorn:
This was actually very easy. Doesn’t need much explanation I think.


maralorn:
@fxrh Be careful only the last commit is actually from this PR, the rest is from #117.

KitsuneRal:
Actually, I do have a small problem with this, but I'm not sure if we can immediately solve it. Basically, it's again about renaming users. As of right now, highlighting is calculated once, upon event arrival, and then is retained as a flag. This PR makes it dynamically calculable, which means that user renames will affect this as well. Probably, it'd better to alter things the other way, pre-calculate more information in the message (and then we might even get away without that room state time machine). This means two containers of messages instead of one in the meantime, yes.

maralorn:
Hm, as you like. I have a very exact idea on how to approach the whole time machine problem, but I didn’t get to it this weekend. I hope that we can merge some branches this week, it’s getting a bit confusing, so that I can approach that next weekend. This is not a promise, but something I’d really like to do.

KitsuneRal:
Let's put this to the future time machine PR then :)

KitsuneRal:
Since `Message` presence started to really stand in the way, I bit the bullet and made essentially the same change as you did but on the up-to-date codebase. Thanks a lot for the effort!

PR #129, opened 2017-03-24
Show messagebox after trying to join an invalid room

### LINKED ISSUE #128: No feedback when trying to join a non-existing room
Running the latest git master. Steps to reproduce:

* Room -> Join Room
* Type a non-existing room name
* Click Ok

Expected behavior: error message
Actual: behavior nothing happens

elvisangelaccio:
If the room the user is trying to join is invalid, show a messagebox
with a warning message instead of doing nothing.

Fixes #128

### DIFF
--- a/client/chatroomwidget.cpp
+++ b/client/chatroomwidget.cpp
@@ -194,14 +194,8 @@ void ChatRoomWidget::sendLine()
     // Commands available without current room
     if( text.startsWith("/join") )
     {
-        QString roomName = text.section(' ', 1, 1, QString::SectionSkipEmpty);
-        if( !roomName.isEmpty() )
-            m_currentConnection->joinRoom( roomName );
-        else
-        {
-            qDebug() << "No arguments for /join, going interactive";
-            emit joinRoomNeedsInteraction();
-        }
+        const QString roomName = text.section(' ', 1, 1, QString::SectionSkipEmpty);
+        emit joinCommandEntered(roomName);
     }
     else // Commands available only in the room context
         if (m_currentRoom)
--- a/client/chatroomwidget.h
+++ b/client/chatroomwidget.h
@@ -49,7 +49,7 @@ class ChatRoomWidget: public QWidget
         void lookAtRoom();
 
     signals:
-        void joinRoomNeedsInteraction();
+        void joinCommandEntered(const QString& roomAlias);
         void showStatusMessage(const QString& message, int timeout);
 
     public slots:
--- a/client/mainwindow.cpp
+++ b/client/mainwindow.cpp
@@ -32,6 +32,7 @@
 #include <QtGui/QMovie>
 #include <QtGui/QCloseEvent>
 
+#include "jobs/joinroomjob.h"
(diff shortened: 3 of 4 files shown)


KitsuneRal:
Thanks for the PR - unfortunately, it's only a half of it. You fell into a trap with Git submodules. The thing is - Quaternion uses libqmatrixclient, and the whole lib folder in its source tree actually belongs to another Git repo. So in order to do such cross-repo changes you have to make two PR's: one for libqmatrixclient, another for Quaternion.

Actually, I was going to do some changes to the related libqmatrixclient code, and just pushed them. So if you don't mind, kindly consider porting your PRs to [this commit](https://github.com/Fxrh/libqmatrixclient/commit/26ab888cf732f768770138f791f1ed7a2bade7f3)

elvisangelaccio:
Yep, I realized my fail the second after I hit the submit button ;)
I will fix the change asap.

KitsuneRal:
Awesome; looks almost good, but it would be good to take care about one more case. In fact, you can join a room by typing /join <room-alias>. The relevant code is in `chatroomwidget.cpp`. Apparently, the same message has to pop out if an incorrect alias is typed there, too. Probably it would be good to shift the whole room-joining code from there to `MainWindow`, renaming `joinRoomNeedsInteraction()` to `joinRoom()` signal so that `MainWindow` creates `JoinRoomJob`, not `ChatRoomWidget`.

KitsuneRal:
Apparently, Clang is not quite impressed with your lambda-fu :) Frankly, I don't really get why would you even need such a construct. Why not just pass a `QString` instead of `const QString &` and assign it from `getText()` if it's empty?

elvisangelaccio:
Weird, I'm building with clang here...
Ok, let's not do it more complicated than it should be.

KitsuneRal:
Almost perfect, but I still would be happy to drop the local variable and just pass a `QString` instead of a `const QString &`.

elvisangelaccio:
Sorry, which local variable are you referring to?

KitsuneRal:
I mean `room`. `joinRoom()` gets a `const QString&` parameter and then you immediately initialize a local variable with the value of this parameter. At the same rate you can just pass to `joinRoom()` a `QString` parameter - in terms of binary code it will be almost the same.
Anyway, I'm now merging this, since it's good enough in its current form; and I'll polish it up with another commit on my side.

elvisangelaccio:
Well `joinRoom(QString room)` will cause a copy. Even though QString is implicitly shared, passying by const reference is still slightly more efficient (and that's also what Qt does in their api).

KitsuneRal:
`QString room = roomAlias;` will equally cause a copy, won't it?

elvisangelaccio:
@KitsuneRal Yep, I was wrong about that. You're right that the binary code will probably be the same. But imho always passing `QString`s by const reference is good practice and also what Qt does in their api.

KitsuneRal:
Ok, that's a valid point. Makes sense from the consistency perspective, too. Using plain `QString` kind of exposes a piece of implementation.

PR #132, opened 2017-03-26
Enable chat line edit only when there is a room

elvisangelaccio:
Otherwise typing into the line edit has no visible effect.

### DIFF
--- a/client/chatroomwidget.cpp
+++ b/client/chatroomwidget.cpp
@@ -91,6 +91,7 @@ ChatRoomWidget::ChatRoomWidget(QWidget* parent)
     m_quickView->setResizeMode(QQuickView::SizeRootObjectToView);
 
     m_chatEdit = new ChatEdit(this);
+    m_chatEdit->setEnabled(false);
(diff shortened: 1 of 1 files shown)


elvisangelaccio:
Actually this change is quite moot, it's the whole `ChatRoomWidget` that should probably be hidden or at least disabled when there is no room yet...

KitsuneRal:
I'm not sure, quite frankly. I can issue commands that don't need a room - say, /logout or /join. Some way of telling that the user ain't gonna get anything would be nice, though. And, frankly, the message box made for the join room case is not the best interface I can imagine. Something less standing in the way would be awesome

elvisangelaccio:
> I'm not sure, quite frankly. I can issue commands that don't need a room - say, /logout or /join.

Right, didn't think about those commands. This change is wrong then.

> Some way of telling that the user ain't gonna get anything would be nice, though. And, frankly, the message box made for the join room case is not the best interface I can imagine. Something less standing in the way would be awesome

In KDE software we use [KMessageWidgets](https://api.kde.org/frameworks/kwidgetsaddons/html/classKMessageWidget.html) for this purpose, but I don't know if introducing such dependency would be acceptable (it's a Qt-only lib, but still).

PR #133, opened 2017-03-26
Add separator between topic and chat widgets

elvisangelaccio:
Otherwise it may not be clear that the date is part of the chat and not
the last line of the topic.


KitsuneRal:
Sorry, actually, I have one more thing here - why this `QFrame` should be a member variable? Why not just make it local?

PR #134, opened 2017-03-30
Make read marker more robust and lighter on resources

KitsuneRal:
Improvements to the code in onMessageShownChanged (former onEventShownChanged):
1. The code doesn't spawn a separate timer for each event/message anymore; a single timer runs at any point in time now, that only tracks the bottommost message
2. There's a single sorted vector for shown messages bookkeeping now, instead of two hash-tables.
3. The code doesn't depend on the model/QML's index - the index may change as new messages get loaded, leading to inaccuracies. The new code also doesn't assume that the read marker only moves down and that it is only updated from the same client.

### DIFF
--- a/client/models/messageeventmodel.cpp
+++ b/client/models/messageeventmodel.cpp
@@ -24,7 +24,6 @@
 #include <QtCore/QDebug>
 
 #include "../message.h"
-#include "../quaternionroom.h"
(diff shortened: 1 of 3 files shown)


KitsuneRal:
Superseded by #135

PR #135, opened 2017-03-30
Detect activity

KitsuneRal:
As there seem to be no reviewers around, I'm encompassing 3 PRs into one to save on bureaucracy; besides, all these actually revolve around the same code. These are:

1. First commit rewrites the shown messages tracking code.
1. The following two move this code out of `MessageEventModel` to `ChatRoomWidget`. The model never seemed to be the right place for such things but we didn't have the chat widget as a context property in QML, so used the model to callback for things.
1. The last two improve `ActivityDetector` - this is basically the code by @maralorn rebased on the newest master and tweaked in some places. Naturally, it doesn't deal with the model anymore, only with the chat widget.

### DIFF
--- a/client/activitydetector.cpp
+++ b/client/activitydetector.cpp
@@ -1,0 +1,73 @@
+/**************************************************************************
+ *                                                                        *
+ * Copyright (C) 2016 Malte Brandy <malte.brandy@maralorn.de>                        *
+ *                                                                        *
+ * This program is free software; you can redistribute it and/or          *
+ * modify it under the terms of the GNU General Public License            *
+ * as published by the Free Software Foundation; either version 3         *
+ * of the License, or (at your option) any later version.                 *
+ *                                                                        *
+ * This program is distributed in the hope that it will be useful,        *
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of         *
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the          *
+ * GNU General Public License for more details.                           *
+ *                                                                        *
+ * You should have received a copy of the GNU General Public License      *
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.  *
+ *                                                                        *
+ **************************************************************************/
+
+#include "activitydetector.h"
+
+#include <QtCore/QDebug>
+
+#include "mainwindow.h"
+#include "chatroomwidget.h"
+
+ActivityDetector::ActivityDetector(QApplication& a, MainWindow& w)
+    : m_app(a)
+    , m_mainWindow(w)
+    , m_enabled(false)
+{
+    const auto chatWidget = w.getChatRoomWidget();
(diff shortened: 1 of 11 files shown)


client/activitydetector.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+#include "mainwindow.h"
+#include "chatroomwidget.h"
+
+ActivityDetector::ActivityDetector(QApplication& a, MainWindow& w)
+    : m_app(a)
+    , m_mainWindow(w)
+    , m_enabled(false)
+{
+    const auto chatWidget = w.getChatRoomWidget();
```
elvisangelaccio: Nitpick: for consistency with the Qt coding style, getters should not start with "get".

client/chatroomwidget.cpp
```diff
@@ -145,6 +142,15 @@ void ChatRoomWidget::setRoom(QuaternionRoom* room)
         m_chatEdit->setText( m_currentRoom->cachedInput() );
         connect( m_currentRoom, &QMatrixClient::Room::typingChanged, this, &ChatRoomWidget::typingChanged );
         connect( m_currentRoom, &QMatrixClient::Room::topicChanged, this, &ChatRoomWidget::topicChanged );
+        connect( m_currentRoom, &QMatrixClient::Room::readMarkerMoved, [=] {
```
elvisangelaccio: lambda functions in a `connect()` may lead to crashes if you don't specify `this` as receiver. See https://medium.com/genymobile/how-c-lambda-expressions-can-improve-your-qt-code-8cd524f4ed9f

client/main.cpp
```diff
@@ -79,8 +54,7 @@ int main( int argc, char* argv[] )
     MainWindow window;
     if( debugEnabled )
         window.enableDebug();
-    ActivityDetector ad(&window);
-    app.installEventFilter(&ad);
+    ActivityDetector ad(app, window); Q_UNUSED(ad);
```
elvisangelaccio: `Q_UNUSED()` doesn't need a comma

KitsuneRal: I still prefer to keep the semicolon, for consistency. Never liked trying to figure which macros in Qt need a semicolon, which don't.

client/activitydetector.cpp
KitsuneRal: Yes, but the current code doesn't have a good naming convention for data members, and coincidentally `chatRoomWidget` is exactly the member variable. Effort to clean this up (and in general impose/document the naming conventions) is welcome.

client/chatroomwidget.cpp
KitsuneRal: That's a very good point, thanks for the link.

KitsuneRal:
Ironically, after adding `this` to `connect()`, Quaternion started crashing on exit :) But that's obviously another bug somewhere. Investigating.
(also yes I know I screwed the ref to lib, will fix it).

PR #136, opened 2017-04-01
Linkify at least some URLs even in plain text messages

### LINKED ISSUE #124: Make URLs clickable
NaChat is doing a remarkably good job at this, maybe borrow from it.

KitsuneRal:
Here's the feature long asked for - URLs in plain text bodies are converted to links now. Surprisingly, I haven't found anything pre-cooked to do that; so I crafted 3 regular expressions for specific situations. Obviously there'll be more things like that, with either more regexps introduced or existing ones extended.

There were several potential ways this can be done:
- Alter `MessageEventContent::TextContent` so that it stores the already linkified URL. The worst option since there's no way to get the pristine plain text in this case.
- Alter `MessageEventContent::TextContent` so that it stores both the original plain text and the linkified version. Better but still buries the representational matter in the backend code. If a client wishes to custom-render some URLs (e.g. for internal navigation to rooms, users etc.) this option immediately falls short.
- Linkify URLs in QML - an obvious choice, given that it's a purely representational matter. Unfortunately, closer inspection shows that QML in itself has no means for that; the only option is to import some external JavaScript linkifier. Which can be a way to go but needs further digging. And it's an external dependency for a pretty much niche case. And there's still a problem with those custom-rendered URLs in the future, since Quaternion doesn't have the whole UI in QML.
- Linkify URLs in `MessageEventModel::data()` - this is how I've done it in the end. There's a clear downside here - since I don't use external linkifier library, I had to invent my own regexp's to find URLs. This is of course a [terrible idea](https://stackoverflow.com/a/21925491) but still better in the situation (see the previous bullet about QML).

Closes #124

### DIFF
--- a/client/models/messageeventmodel.cpp
+++ b/client/models/messageeventmodel.cpp
@@ -21,7 +21,6 @@
 
 #include <QtCore/QTimerEvent>
 #include <QtCore/QSettings>
-#include <QtCore/QRegularExpression>
 #include <QtCore/QDebug>
 
 #include "../message.h"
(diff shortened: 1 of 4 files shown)


client/models/messageeventmodel.cpp
```diff
@@ -112,6 +113,44 @@ int MessageEventModel::rowCount(const QModelIndex& parent) const
     return m_currentRoom->messages().count();
 }
 
+void linkifyUrls(QString& text)
+{
+    static const auto RegExpOptions =
+        QRegularExpression::CaseInsensitiveOption
+#if (QT_VERSION >= QT_VERSION_CHECK(5, 4, 0))
```
elvisangelaccio: Do we really need to support old Qt versions? Everyone should be using at least Qt 5.6.x by now

client/models/messageeventmodel.cpp
```diff
@@ -112,6 +113,44 @@ int MessageEventModel::rowCount(const QModelIndex& parent) const
...
+        | QRegularExpression::OptimizeOnFirstUsageOption
+#endif
+        | QRegularExpression::UseUnicodePropertiesOption;
+    static const QRegularExpression urlDetectorMail {
+        "(^|\\s)(" // Criteria to match the beginning of the URL
```
elvisangelaccio: QStringLiteral? Same for the other literals below.

client/models/messageeventmodel.cpp
KitsuneRal: We currently have a convention to support the Qt version used in the current stable Debian. So far this is 5.3.2, to be fair; we still support 5.2.1 because somebody asked about it in the past and it's virtually no-effort compared to 5.3.2. Once stretch becomes stable, and unless there are other obstacles, we'll jump to Qt 5.7.

client/models/messageeventmodel.cpp
KitsuneRal: Thanks, makes sense.

elvisangelaccio:
Tested the PR locally and it works fine with links in the chat messages. Is it possible to do the same with the links the room topic?

elvisangelaccio:
hmm actually I just found a small issue, "version numbers" are also linkified, e.g. `0.20` is detected as if it were a URL.

KitsuneRal:
Moreover, timestamps (like 09:00) were linkified. All fixed.

elvisangelaccio:
> Moreover, timestamps (like 09:00) were linkified. All fixed.

What about timestamps like `10:30:59`? I still get those linkified

KitsuneRal:
Right, not really fixed :-/

elvisangelaccio:
Just noticed that strings like `foo:bar` are also linkified. It seems nachat gets this feature right, maybe we can investigate how they do it?

KitsuneRal:
Actually, I've already found a culprit - I went too far with assuming things like urn:things or (closer to the point) turn:servers. I should probably require at least one dot in the absolute URL.

JFYI, here's [the Nachat regexp](https://github.com/Ralith/nachat/blob/master/src/TimelineView.cpp#L75). I'm afraid Nachat is not so good at linkifying either, only it's too narrow, while my current regexp is too broad.

KitsuneRal:
Ok, case in point: do we want to support URIs like about:config, for example, or turn:localhost? Or these are edge cases and those who wish to open them can do that by copy-pasting the URI where necessary?

KitsuneRal:
As discussed in [#quaternion](https://matrix.to/#/#quaternion:matrix.org), play safe and keep all those cases plain.

elvisangelaccio:
Looks good to me now :)

KitsuneRal:
Thanks, I really didn't think of using the line edit in such a way. We should probably switch to using the rich-text edit control for message entry as well (getting bold/italic/underscore formatting almost for free along the way). On the other hand, `\n`'s in plaintext messages have to be converted to `<br/>`'s anyway. Will follow up directly in master with this.

elvisangelaccio:
> We should probably switch to using the rich-text edit control for message entry as well (getting bold/italic/underscore formatting almost for free along the way)

I'm actually playing with a `QTextEdit` instead of the `QLineEdit` currently in use (in `ChatEdit`). Is this what you had in mind?

KitsuneRal:
Yes, exactly! One thing that concerns me is how much discoverable those rich-text options can be without a toolbar. I'd hate a toolbar as a waste of space; I'd rather see some kind of a quick cheat sheet with hotkeys to do certain formatting that would be accessible with some button next to the input, or something like that.

KitsuneRal:
Yes, exactly! One thing that concerns me is how much discoverable those rich-text options can be without a toolbar. I'd hate a toolbar as a waste of space; I'd rather see some kind of a quick cheat sheet with hotkeys to do certain formatting that would be accessible with some button next to the input, or something like that.

elvisangelaccio:
Cool, will open a PR soon (excluding the formatting, where I agree that a toolbar would take too much space) :)

KitsuneRal:
Looking forward to it. Especially interested how you dealt with completion.

PR #137, opened 2017-04-04
Improve the tooltip of rooms

elvisangelaccio:
The display name is already shown in the `DisplayRole`. The canonical
alias instead can be useful to disambiguate rooms with the same display
name.


KitsuneRal:
How about showing some more useful information in the tooltip? Room id maybe, the number of participants, the number of highlighted messages?

PR #139, opened 2017-04-13
Improve chat edit widget

elvisangelaccio:
Use KChatEdit rather than a custom QLineEdit. KChatEdit is designed as
general-purpose input widget for chat applications, not Matrix-specific.
It has built-in support for history. Tab completion is achieved by
installing a simple event filter.

Rich-text is disabled, for now. We need support for it in the backend
library first.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,6 +72,7 @@ set(quaternion_SRCS
     client/mainwindow.cpp
     client/roomlistdock.cpp
     client/userlistdock.cpp
+    client/kchatedit.cpp
     client/chatroomwidget.cpp
     client/systemtray.cpp
     client/models/messageeventmodel.cpp
--- a/client/chatroomwidget.cpp
+++ b/client/chatroomwidget.cpp
@@ -38,6 +38,31 @@
 #include "models/messageeventmodel.h"
 #include "imageprovider.h"
 
+class ChatEdit : public KChatEdit
+{
+public:
+    ChatEdit(ChatRoomWidget* c, QWidget* parent = nullptr);
+protected:
+    void keyPressEvent(QKeyEvent* event) override;
+private:
+    ChatRoomWidget* m_chatRoomWidget;
+};
+
+ChatEdit::ChatEdit(ChatRoomWidget* c, QWidget* parent)
+    : KChatEdit(parent)
+    , m_chatRoomWidget(c) {};
+
+void ChatEdit::keyPressEvent(QKeyEvent* event)
+{
+    if (event->key() == Qt::Key_Tab) {
+        m_chatRoomWidget->triggerCompletion();
+        return;
+    }
+
+    m_chatRoomWidget->cancelCompletion();
+    KChatEdit::keyPressEvent(event);
+}
+
 ChatRoomWidget::ChatRoomWidget(QWidget* parent)
     : QWidget(parent)
     , m_currentRoom(nullptr)
@@ -65,10 +90,9 @@ ChatRoomWidget::ChatRoomWidget(QWidget* parent)
     m_quickView->setSource(QUrl("qrc:///qml/chat.qml"));
     m_quickView->setResizeMode(QQuickView::SizeRootObjectToView);
 
-    m_chatEdit = new KChatEdit(this);
+    m_chatEdit = new ChatEdit(this);
     m_chatEdit->setPlaceholderText(tr("Send a message (unencrypted)..."));
     m_chatEdit->setAcceptRichText(false);
-    m_chatEdit->installEventFilter(this);
     connect( m_chatEdit, &KChatEdit::inputChanged, this, &ChatRoomWidget::sendLine );
 
     m_currentlyTyping = new QLabel();
@@ -394,21 +418,6 @@ void ChatRoomWidget::markShownAsRead()
         Q_ASSERT( iter != m_currentRoom->timelineEdge() );
         m_currentRoom->markMessagesAsRead((*iter)->id());
     }
-}
-
-bool ChatRoomWidget::eventFilter(QObject *object, QEvent *event)
-{
-    if (object != m_chatEdit || event->type() != QEvent::KeyPress)
-        return QWidget::eventFilter(object, event);
-
-    auto keyEvent = static_cast<QKeyEvent*>(event);
-    if (keyEvent->key() == Qt::Key_Tab) {
-        emit triggerCompletion();
-        return true;
-    }
-
-    emit cancelCompletion();
-    return false;
 }
 
 bool ChatRoomWidget::pendingMarkRead() const
--- a/client/chatroomwidget.cpp
+++ b/client/chatroomwidget.cpp
@@ -18,6 +18,7 @@
  **************************************************************************/
 
 #include "chatroomwidget.h"
+#include "kchatedit.h"
 
 #include <QtCore/QDebug>
 #include <QtWidgets/QLineEdit>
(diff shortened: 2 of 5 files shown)


elvisangelaccio:
Note: I'm planning to submit `KChatEdit` as part of the [KWidgetsAddons library](https://api.kde.org/frameworks/index.html#sg-tier_1) and would be awesome if in the long-term Quaternion could link against this library. But for now I'm just copying the class sources in the Quaternion repo.

KitsuneRal:
Well, I'm not sure what to say here yet. Just putting some thoughts and a bit of history:

- The reason why I was thinking about `QTextEdit` was exactly rich text editing. This PR is not relevant to rich text editing but that's ok.
- Out of features that `KChatEdit` brings on top of `QTextEdit` there seems to be the only one, it's history maintenance - and this, apparently is what this PR is about in functional terms. Maybe it's just me, but I only see the use of history in ability to edit my former messages. There's a problem with messages editing in case of Matrix, though: as of right now, AFAIK, it is not implemented even on the server side, let alone fixed in the spec. Which leads to the dubious UX with, in particular, Riot, which has message history but an attempt to pick an older message, edit it and send brings me just a duplicate of the message I've sent before. Add to that that Key_Up is used to move the cursor through multiline message and you have a bit of a mess. Just try to type a few lines in Riot and then figure out when Key_Up would replace _the whole message you've just typed_ with an older message you don't need now in the first place. So I'm opposed to adding message history feature that makes UX trickier and doesn't bring much added value (=message editing).
- The trouble that we've already stumbled upon with another (equally isolated from the rest of KDE) K-class, `KJob` is that people look at it, say "aw, this thing needs KDE" and turn away. Specifically `KJob` was, moreover, rather standing in the way with its restrictive interface instead of facilitating background requests business. In the end I took those less than 100 lines from `KJob` to `BaseJob` that were really useful, and thrown out the dependency on KCoreAddons. And a bit counter-intuitively things became easier: besides more control of the class interface, I didn't need to special-case Windows that never has KCoreAddons (so we used it as a git submodule) from KDE that always has it system-wide and from other Linux environments that may or may not have KCoreAddons (so we made a compilation switch in CMakeLists.txt). Which taught me a lesson that every time I take an external dependency I also take restrictions of that dependency, an additional moving part in the build configuration, and a possible frown from people who are quick to make far-flying conclusions from a quick grasp on README.md.

Once written all this, I guess I've made my mind. Basically, as soon as history management is the only `KChatEdit`'s differentiator, I'd like to hold off until Matrix is able to support the respective use case (message editing) from the server side. That is, unless you give me other use cases where message history is useful. And I'd also prefer to have some bigger value from an external dependency than just history management (which very much reminds me of a story with `KJob`).

elvisangelaccio:
Wow, long comment ;)

> Out of features that KChatEdit brings on top of QTextEdit there seems to be the only one, it's history maintenance

There is also some default size handling, but yep that's minor.

> Maybe it's just me, but I only see the use of history in ability to edit my former messages.

Personal use-case: I send very often to the ` @appservice-irc:matrix.org` bot either the `!cmd AWAY` message or the `!cmd AWAY <away_reason>` message. With history I would just type Up and adjust the previous message, without history I need to type the whole `!cmd AWAY ...`, which is annoying.

And what about typos? I can just press Up, fix my typo and send a new message. I don't really care to edit the message I already sent.

> Which leads to the dubious UX with, in particular, Riot, which has message history but an attempt to pick an older message, edit it and send brings me just a duplicate of the message I've sent before

Not sure how to reproduce this, in Riot if I press the Up button and edit the older message, the new message I send is different.

> Just try to type a few lines in Riot and then figure out when Key_Up would replace the whole message you've just typed with an older message you don't need now in the first place.

You can press Key_Down and get back the draft you were typing. I don't think this is a big deal.

> The trouble that we've already stumbled upon with another (equally isolated from the rest of KDE) K-class, KJob is that people look at it, say "aw, this thing needs KDE" and turn away. Specifically KJob was, moreover, rather standing in the way with its restrictive interface instead of facilitating background requests business. In the end I took those less than 100 lines from KJob to BaseJob that were really useful, and thrown out the dependency on KCoreAddons. And a bit counter-intuitively things became easier: besides more control of the class interface, I didn't need to special-case Windows that never has KCoreAddons (so we used it as a git submodule) from KDE that always has it system-wide and from other Linux environments that may or may not have KCoreAddons (so we made a compilation switch in CMakeLists.txt). Which taught me a lesson that every time I take an external dependency I also take restrictions of that dependency, an additional moving part in the build configuration, and a possible frown from people who are quick to make far-flying conclusions from a quick grasp on README.md.

I was not aware of this. And you are not the only ones btw, for example https://github.com/frankosterfeld/qtkeychain does exactly the same.

I can see the advantages of forking, but imho the cons are worse. You are duplicating code which you need to maintain, you are not getting back bugfixes, you are losing the unit tests, etc.
If KJob is hard to use on Windows, then we should fix that. Have you ever tried to contact the KDE community and report the issues you faced with the integration of KCoreAddons?

Also, the "needs KDE" scare argument is moot. KDE is a community, not a desktop. The desktop is called Plasma and has nothing to do with Qt-only libraries such as KCoreAddons or KWidgetAddons. If people run away because they see a K prefix that's their problem, you should not be afraid of using existing and well-tested libraries, which could improve Quaternion, because of this.

KitsuneRal:
Sorry for another long one. I guess that should be the last one, and I'll think further on the matter.
Ok, the bots case is clear. Just double-checked Riot and seems that something changed over the last couple of months - most likely I was hitting some bug before. Now the UX is pretty straightforward. I previously stumbled over a situation when Down didn't get me the half-written message anymore - it was basically lost. And I still can't get why it should have been the same key as the one used for an entirely different purpose (of course I know about the command line but still) - anyway, this is ranting already.
Now to the serious stuff. Basically, I also thought that although it's a trade-off, it's better to use than to fork. Apparently, it depends on how well targets of the external dep are aligned with ours. As far as I remember my issue with `KJob`, the semantics of its signals didn't go well with our stuff; in particular, the logic behind jobs abandoning and timeouts was basically preventing reuse of a job for retries, e.g. At the same time, the whole download progress tracking infrastructure that `KJob` has was entirely unused because our "jobs" (actually, network requests) were not really suitable for uniform progress tracking. `KJob` was most likely good for cases like a file manager or a web browser but for instant messaging it just didn't fit well. I could work with upstream trying to fix it but I'm not in the KDE community, and more importantly, the amount of code to advocate for was unreasonable. Cm'on, 100 very straightforward lines, most of which are documentation...
Then there's still that headache with various components that can be installed or not in the system. Cross-platform development is especially sensitive to that. The more libraries you have around, the more your build configuration becomes a hell of compilation switches and autodetection what you have around and what you don't. So I prefer to link to bigger parts, like Qt::Gui or Boost.Network (just an example). That became one of problems for Nachat, btw - its dependencies are not so easy to configure and build, in particular, on Windows. People (including me) just gave up to invest so much upfront.
As for the K-flamewar - I actually know very well how KDE is organized these days but I can't say so of many people around me. In general people are ignorant and tend to oversimplify things. Basically, I had to dismiss an improper question "Is it a Qt application or a KDE application?" several times and I thought I didn't want to talk about such things. And given that Qt provided almost everything necessary, I was sticking to the "pure-Qt" tag. But that's actually less important comparing to the roadmap/profile and maintenance issues mentioned above.

KitsuneRal:
Ok, upon some thinking - let's get it in. That code for sizing isn't very simple, actually; and my quick attempt to transfer from `QLineEdit` to `QTextEdit` didn't go out as easy as I thought, either. So be it :)

elvisangelaccio:
>As far as I remember my issue with KJob, the semantics of its signals didn't go well with our stuff; in particular, the logic behind jobs abandoning and timeouts was basically preventing reuse of a job for retries, e.g. At the same time, the whole download progress tracking infrastructure that KJob has was entirely unused because our "jobs" (actually, network requests) were not really suitable for uniform progress tracking. KJob was most likely good for cases like a file manager or a web browser but for instant messaging it just didn't fit well. I could work with upstream trying to fix it but I'm not in the KDE community, and more importantly, the amount of code to advocate for was unreasonable. Cm'on, 100 very straightforward lines, most of which are documentation...

I see. I'm not saying that we should switch back to KJob, what's done is done. My point is that if N people are extracting those 100 lines of code over and over, they are just doing the same work N times. If someone fixes KJob once and for all instead, the next guy who tries to use KJob probably won't have to do that work the N+1th time ;)

> Cross-platform development is especially sensitive to that. The more libraries you have around, the more your build configuration becomes a hell of compilation switches and autodetection what you have around and what you don't. So I prefer to link to bigger parts, like Qt::Gui or Boost.Network (just an example). That became one of problems for Nachat, btw - its dependencies are not so easy to configure and build, in particular, on Windows. People (including me) just gave up to invest so much upfront.
As for the K-flamewar - I actually know very well how KDE is organized these days but I can't say so of many people around me. In general people are ignorant and tend to oversimplify things. Basically, I had to dismiss an improper question "Is it a Qt application or a KDE application?" several times and I thought I didn't want to talk about such things. And given that Qt provided almost everything necessary, I was sticking to the "pure-Qt" tag. But that's actually less important comparing to the roadmap/profile and maintenance issues mentioned above.

I'd argue "that's what cmake is for", it is supposed to hide the ugliness of cross-platform deployment so that in your code you can just focus on features. But ok, this is a discussion for another time :)
I'm still going to submit `KChatEdit` for review in KWidgetsAddons, because I think it can be useful for more than one application (and I also have unit tests for it). If and how we want to link quaternion against KWidgetsAddons, can be discussed later.

elvisangelaccio:
Cool, thanks for merging!

PR #141, opened 2017-04-15
Give focus to chat edit widget when changing rooms

elvisangelaccio:
And also move the cursor to the end of the text.


elvisangelaccio:
hmm actually I'm wondering if the chat edit should have its own api to set the text and automatically move the cursor to the end. This is what QLineEdit does, while QTextEdit doesn't.

elvisangelaccio:
> hmm actually I'm wondering if the chat edit should have its own api to set the text and automatically move the cursor to the end. This is what QLineEdit::setText() does, while QTextEdit::setText() doesn't.

This is also reason, by the way, why the cursor no longer moves to the end of the completion text.

KitsuneRal:
I'm now reworking the whole completion text so that it works with rich text. Setting the cursor to the end occurs very naturally in it.

Issue #142, opened 2017-04-17
KChatEdit clears the entry before any attempt to check/send the message


KitsuneRal:
The issue is two-fold:
- End-user experience: we already have a check for a lone / in the beginning of the message. Since `KChatEdit` clears the entry before emitting `inputChanged` signal (which is incorrect, to my mind), the end-user sees a status message and an already cleared line. Yes he/she can push Up and get his input but that's just not right, the line should be there until actually submitted to the server.
- Developer experience: I'm trying to adapt `ChatRoomWidget::sendLine()` to rich text messages. What I need for that is access to the original representation (not only HTML or only plain text) so that I could navigate through the underlying `QTextDocument` and convert/validate it. Once rich-text input becomes possible, I really don't want to mess with HTML tags in that place. But there's no clean way to do it: the history remembers only HTML /or/ plaintext representation, and filling in a `QTextDocument` back from history is plain stupid.

What I think would be right is to keep the entry intact until the end of `ChatRoomWidget::sendLine()` and once we're done with it, call some kind of `KChatEdit::save()` to push the line to the room history.

elvisangelaccio:
Makes sense. The first issue should be easy to fix, not sure about the second: what about making the history a list QTextDocument instead of a QStringList?

PR #145, opened 2017-04-17
Port to KChatEdit::saveInput()

elvisangelaccio:
Pressing Return now only emits the inputChanged() signal, but neither
clears the text edit nor saves the input in the history.
The new saveInput() method explicitly takes care of these tasks.

See issue #142.


elvisangelaccio:
@KitsuneRal Commit 5ed60fd is probably orthogonal to this PR, but I pushed it anyway.

KitsuneRal:
Uhm, it's actually relevant to one of my requests; but why go through `document()` if `QTextEdit` has the same method anyway?

elvisangelaccio:
Superseded by #147.

PR #147, opened 2017-04-17
Improve chatedit widget

### LINKED ISSUE #142: KChatEdit clears the entry before any attempt to check/send the message

elvisangelaccio:
Previously the chat edit widget provided only QString objects, containing
either plain text or html. Now the api is much more flexible, as the caller
can choose what to do with the QTextDocument.

Another change in behavior is that pressing Return no longer clears the
input and puts it in the history. A returnPressed() signal notifies when
Return is pressed, and a new saveInput() method can be used to
explicitly accept an input and store it in the history.

Fixes #142

### DIFF
--- a/client/chatroomwidget.h
+++ b/client/chatroomwidget.h
@@ -32,6 +32,7 @@ class QQuickView;
 class QListView;
 class QLineEdit;
 class QLabel;
+class QTextDocument;
 
 class ChatRoomWidget: public QWidget
 {
@@ -63,7 +64,7 @@ class ChatRoomWidget: public QWidget
         void timerEvent(QTimerEvent* event) override;
 
     private slots:
-        void sendLine();
+        void sendInput();
 
     private:
         MessageEventModel* m_messageModel;
@@ -91,7 +92,7 @@ class ChatRoomWidget: public QWidget
         timeline_index_t indexToMaybeRead;
         QBasicTimer maybeReadTimer;
         bool readMarkerOnScreen;
-        QMap<QuaternionRoom*, QStringList> roomHistories;
+        QMap<QuaternionRoom*, QList<QTextDocument*>> roomHistories;
 
         void reStartShownTimer();
 };
--- a/client/chatroomwidget.h
+++ b/client/chatroomwidget.h
@@ -92,7 +92,7 @@ class ChatRoomWidget: public QWidget
         timeline_index_t indexToMaybeRead;
         QBasicTimer maybeReadTimer;
         bool readMarkerOnScreen;
-        QMap<QuaternionRoom*, QList<QTextDocument*>> roomHistories;
+        QMap<QuaternionRoom*, QVector<QTextDocument*>> roomHistories;
 
         void reStartShownTimer();
 };
--- a/client/kchatedit.cpp
+++ b/client/kchatedit.cpp
@@ -37,7 +37,7 @@ public:
     void saveInput();
 
     KChatEdit *q;
-    QList<QTextDocument*> history;
+    QVector<QTextDocument*> history;
     int index;
     int maxHistorySize;
 };
@@ -92,7 +92,7 @@ void KChatEdit::KChatEditPrivate::saveInput()
     }
 
     const QString input = q->acceptRichText() ? q->toHtml() : q->toPlainText();
-    const QString latestInput = q->acceptRichText() ? q->input()->toHtml() : q->input()->toPlainText();
+    const QString latestInput = q->acceptRichText() ? q->savedInput()->toHtml() : q->savedInput()->toPlainText();
     // Only save input if different from the latest one.
     if (input != latestInput) {
         // Replace empty placeholder with the new input.
@@ -105,7 +105,7 @@ void KChatEdit::KChatEditPrivate::saveInput()
     }
 
     index = history.size() - 1;
-    emit q->inputChanged();
+    emit q->savedInputChanged();
     q->clear();
 }
 
@@ -121,7 +121,7 @@ KChatEdit::~KChatEdit()
 {
 }
 
-QTextDocument* KChatEdit::input() const
+QTextDocument* KChatEdit::savedInput() const
 {
     if (d->history.size() >= 2)
         return d->history.at(d->history.size() - 2);
@@ -135,12 +135,12 @@ void KChatEdit::saveInput()
     d->saveInput();
 }
 
-QList<QTextDocument*> KChatEdit::history() const
+QVector<QTextDocument*> KChatEdit::history() const
 {
     return d->history;
 }
 
-void KChatEdit::setHistory(const QList<QTextDocument*> &history)
+void KChatEdit::setHistory(const QVector<QTextDocument*> &history)
 {
     d->history = history;
     if (history.isEmpty() || !history.last()->isEmpty()) {
--- a/client/kchatedit.cpp
+++ b/client/kchatedit.cpp
@@ -31,12 +31,13 @@ public:
     {}
 
     void init();
+    void emitReturnPressed();
     void rewindHistory();
     void forwardHistory();
     void saveInput();
 
     KChatEdit *q;
-    QStringList history;
+    QList<QTextDocument*> history;
     int index;
     int maxHistorySize;
 };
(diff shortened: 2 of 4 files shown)


client/kchatedit.cpp
```diff
@@ -31,72 +31,82 @@ class KChatEdit::KChatEditPrivate
...
 
         if(history.size() > maxHistorySize) {
             history.removeFirst();
         }
     }
 
     index = history.size() - 1;
-
+    emit q->inputChanged();
     q->clear();
```
KitsuneRal: This looks clumsy. You first emit a signal that the input has changed (a casual reader would assume you mean that text on the screen), only to immediately clear it afterwards. You had to rename the signal to `inputAboutToBeChanged()` according to Qt conventions in that case. But it's actually more complicated, read on.
To begin with, we have to better distinguish between the actual input (the text on the screen) and its base document in the history. The point where the confusion manifests most is in `sendInput()` - you have `input`-the-variable that means not the same as `input`-the-property but the text on the screen, and `latestInput` that is taken from the `input`-the-property.
Probably it's better to rename the property to `inputSource` or `inputBase` or something like that. And I'm not sure how much it's supposed to be in the public interface - I might just leave one accessor method, to use in `sendInput()`. Then `saveInput()` would genuinely mean "save what user entered in the history", and if we ever want to expose the history manipulation interface, this would be accompanied by some `historyChanged()` signal (but I think it would be over-engineering to introduce it right away).

client/kchatedit.cpp
```diff
@@ -111,21 +121,30 @@ KChatEdit::~KChatEdit()
 {
 }
 
-QString KChatEdit::input() const
+QTextDocument* KChatEdit::input() const
```
KitsuneRal: As mentioned above, this doesn't really look like "user input" (which I first implied when reading the code) but rather a base of this input in history. Let's rename it.

client/kchatedit.cpp
```diff
@@ -200,7 +223,7 @@ void KChatEdit::keyPressEvent(QKeyEvent *event)
         }
         break;
     case Qt::Key_Down:
-        if (!document()->isEmpty() && !textCursor().movePosition(QTextCursor::Down)) {
+        if (!textCursor().movePosition(QTextCursor::Down)) {
```
KitsuneRal: What if I took something from the middle of the history, edited it and then accidentally switched to another history entry? My edits gone?

client/kchatedit.h
```diff
@@ -33,44 +33,52 @@
...
+    Q_PROPERTY(QList<QTextDocument*> history READ history WRITE setHistory)
     Q_PROPERTY(int maxHistorySize READ maxHistorySize WRITE setMaxHistorySize)
 
 public:
     explicit KChatEdit(QWidget *parent = nullptr);
     virtual ~KChatEdit();
 
     /**
-     * The latest input text that the user provided by pressing the Return key.
+     * The latest input text typed by the user.
```
KitsuneRal: That's not quite true. As discussed above, actually this is the text from history used as a base for the currently edited text.

client/kchatedit.cpp
elvisangelaccio: > You had to rename the signal to inputAboutToBeChanged() according to Qt conventions in that case.

Why? By the time we emit that signal, the input property *has* changed.
I'd rather rename the local variables in this function if they are confusing.

client/kchatedit.cpp
elvisangelaccio: Not sure how should be renamed. Either this method is useful, or it's not and should be removed rather than renamed.

The main idea of the `input()` method was to retrieve the text after it got cleread. So basically:

* connect to `returnPressed()`
* if everything is ok, call `saveInput()` to "commit" this input
* at any later time, call `input()` if you need the very latest input typed by the user.

client/kchatedit.cpp
elvisangelaccio: Nope, changes are saved before switching items.

client/chatroomwidget.h
```diff
@@ -91,7 +92,7 @@ class ChatRoomWidget: public QWidget
         timeline_index_t indexToMaybeRead;
         QBasicTimer maybeReadTimer;
         bool readMarkerOnScreen;
-        QMap<QuaternionRoom*, QStringList> roomHistories;
+        QMap<QuaternionRoom*, QList<QTextDocument*>> roomHistories;
```
elvisangelaccio: Isn't `QVector` worse if we need to remove the first element? (shift all the elements to keep the contiguous structure)

About `QCache`, what would you use as key for the history items?

client/kchatedit.cpp
KitsuneRal: Well, apparently it's useful, because you use it :) It's just that "the very latest input typed by the user", the way I see it is exactly the input in the text box. Once saved, the input becomes history. Well, maybe `savedInput()` then?

client/chatroomwidget.h
KitsuneRal: As for `QVector` inefficiency for the shifting case - you either move a pack of elements once you remove the first one, or you allocate a new piece of memory once in a while to fit the newly added elements (and `QList` doesn't deallocate memory that is no more used, unless you explicitly call `shrink()`, as far as I remember). I prefer the first inefficiency; `malloc()`, even once in a while, is more expensive than `memcpy()`.
In terms of abstract data types, the best thing for this would probably be a circular buffer. There's no such thing as a container but the implementation is trivial on top of virtually any random-access container you want. I suppose, `QContiguousCache` is closer to this than anything.
For `QCache` - probably HTML representations of the `QTextDocument` would go for keys. That would help with saving duplicate messages in the history at the same time. But it looks like `QContiguousCache` fits the purpose much better - it doesn't store items in a map, rather just indexes them.

client/kchatedit.cpp
KitsuneRal: Oh, right, sorry, missed that in `rewindHistory()` and `forwardHistory()`.

client/kchatedit.cpp
elvisangelaccio: Ok then, `savedInput()` sounds good indeed, I can see how it can be less ambiguous :)

client/chatroomwidget.h
elvisangelaccio: Honestly there is a risk of premature optimization here. I'd go with the most readable and simpler implementation. Let's go with `QVector` and we should be fine, I think?

Also, I kinda agree with you that the history manipulation should not be part of the public api. Instead of using a single instance of `ChatEdit` and maintaining a map between rooms and histories, we could use one `ChatEdit` instance for each room. What do you think?

KitsuneRal: Ok, let it be `QVector`. As for multiple `ChatEdit` objects - probably let's keep it single for the moment to keep the PR small. But actually, I already think about going further and making the whole `ChatRoomWidget` bound to a certain room. As discussed in #quaternion, at least one user runs several Quaternion instances to have several rooms open. I believe I've seen some way to easily detach tabs/pages from a Qt tab view control; we might use this to detach rooms into separate windows if a user wishes so. But we might need to do some lazy instantiation to avoid scaling problems with huge accounts with a thousand of rooms.

elvisangelaccio:
@KitsuneRal I'm going to use this PR only for the changes to the public API, so that you can go on with your refactoring. Everything else (history trimming etc.) can be done later with another PR :)

PR #149, opened 2017-04-22
Don't lose focus when clicking a room twice

elvisangelaccio:
Without this setFocus() call, the focus would go to the room view
widget.


KitsuneRal:
Accepting this under assumption that if we want to make some form of "room peeking" we can easily refactor the code so that `setRoom()` sets the focus conditionally.

PR #151, opened 2017-04-30
Chat edit refactored

KitsuneRal:
The refactoring moves the tab completion code to `ChatEdit` where it naturally belongs and, since `ChatEdit` becomes quite a bit of almost self-contained functionality, it gets its own pair of files. Oh, and notably - the tab completion code now uses `QTextCursor` instead of indices, which allows to input rich text without breaking completion. Bonus: fancy underlining of the currently-completed piece.

### DIFF
--- a/client/chatroomwidget.cpp
+++ b/client/chatroomwidget.cpp
@@ -78,7 +78,7 @@ ChatRoomWidget::ChatRoomWidget(QWidget* parent)
             [=](const QStringList& matches, int pos)
             {
                 m_currentlyTyping->setText(
-                    QStringLiteral("<i>Tab Completion (next: %1)</i>")
+                    tr("<i>Tab Completion (next: %1)</i>")
                     .arg( QStringList(matches.mid(pos, 5)).join(", ") ) );
             });
     connect(m_chatEdit, &ChatEdit::cancelledCompletion,
@@ -168,7 +168,7 @@ void ChatRoomWidget::typingChanged()
     {
         typingNames << m_currentRoom->roomMembername(user);
     }
-    m_currentlyTyping->setText(QStringLiteral("<i>Currently typing: %1</i>")
+    m_currentlyTyping->setText(tr("<i>Currently typing: %1</i>")
                                .arg( typingNames.join(", ") ) );
 }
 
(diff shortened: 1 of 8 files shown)


client/chatroomwidget.cpp
```diff
@@ -90,21 +68,21 @@ ChatRoomWidget::ChatRoomWidget(QWidget* parent)
...
-    m_topicLabel->setWordWrap(true);
-    m_topicLabel->setTextInteractionFlags(Qt::TextBrowserInteraction
-                                          |Qt::TextSelectableByKeyboard);
-    m_topicLabel->setOpenExternalLinks(true);
-
+    connect(m_chatEdit, &ChatEdit::proposedCompletion, this,
+            [=](const QStringList& matches, int pos)
+            {
+                m_currentlyTyping->setText(
+                    QStringLiteral("<i>Tab Completion (next: %1)</i>")
```
elvisangelaccio: `tr()` ?

client/chatroomwidget.cpp
```diff
@@ -184,23 +158,30 @@ void ChatRoomWidget::setConnection(QMatrixClient::Connection* connection)
...
         return;
     }
     QStringList typingNames;
-    for( QMatrixClient::User* user: typing )
+    for(auto user: m_currentRoom->usersTyping())
     {
         typingNames << m_currentRoom->roomMembername(user);
     }
-    m_currentlyTyping->setText( QString("<i>Currently typing: %1</i>").arg( typingNames.join(", ") ) );
+    m_currentlyTyping->setText(QStringLiteral("<i>Currently typing: %1</i>")
```
elvisangelaccio: `tr()` here as well

client/chatroomwidget.cpp
```diff
@@ -184,23 +158,30 @@ void ChatRoomWidget::setConnection(QMatrixClient::Connection* connection)
...
+                               .arg( typingNames.join(", ") ) );
 }
 
 void ChatRoomWidget::topicChanged()
 {
-    m_topicLabel->setText( m_currentRoom->prettyTopic() );
+    if (m_currentRoom)
+    {
+        auto topic = m_currentRoom->topic();
+        m_topicLabel->setText(topic.isEmpty() ? QStringLiteral("(no topic)") :
```
elvisangelaccio: and here.

KitsuneRal:
Right you are, thanks for noticing.

elvisangelaccio:
+1, tested and works nicely.

KitsuneRal:
So be it, then.

PR #152, opened 2017-05-02
Add build script.

sbts:
build_quaternion.sh allows a single command to build (and optionally update the local git repo) quaternion

### DIFF
--- a/build_quaternion.sh
+++ b/build_quaternion.sh
@@ -1,0 +1,92 @@
+#!/bin/bash
+
+SELF="$0"
+ARGS="$@"
+
+DIE() {
+    E=$1; shift
+    cat <<-EOF
+	
+	============================
+	== FAIL: $@
+	============================
+	
+	EOF
+    exit $E
+}
+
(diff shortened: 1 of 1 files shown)


KitsuneRal:
What you're trying to do is more than halfway of packaging the application. Maybe make a full-fledged debian/ directory then? Somebody else would bring in the .spec file for Fedora, and will be good to go on 80% of the user base (Arch is already covered externally, by the way).

KitsuneRal:
...and I promise to have some build counter so that the build packages could be properly upgraded then :)

elsieholmes:
>  .spec file for Fedora

And for OpenSUSE. These are **different** distros. Plus, you can build packages for different distros on SUSE's OBS.

KitsuneRal:
We now have a Flatpak, the package is submitted for inclusion into Debian, and the work on packaging for OBS is ongoing - see https://github.com/QMatrixClient/Quaternion/issues/211. Thanks for your help anyway!

PR #154, opened 2017-05-03
Add support for flatpak

elvisangelaccio:
Flatpak is a Linux tool that allows cross-distribution packages. This
commit adds the JSON manifest file necessary to locally build a flatpak package.

Qt is picked from the flatpak runtime provided by KDE. A couple of
scripts contain the necessary flatpak commands to build and install the
flatpak package.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,23 @@ There's no official packaging at the moment. Some links to unofficial packages f
 Arch Linux: https://aur.archlinux.org/packages/quaternion/
 Windows: automatic builds are packaged upon every commit (CI) at https://ci.appveyor.com/project/Fxrh/quaternion (go to "Jobs", select the job for your architecture, then "Artifacts")
 
+## Flatpak
+
+If your run Linux and your distribution supports flatpak, you can easily build and install Quaternion as flatpak package:
+
+```
+git clone https://github.com/Fxrh/Quaternion.git --recursive
+cd Quaternion/flatpak
+./setup_repo.sh
+./build.sh
+flatpak --user install quaternion-nightly com.github.quaternion
+```
+Whenever you want to update your Quaternion package, just run again the `build.sh` script and then use:
+
+```
+flatpak --user update
+```
+
 ## Troubleshooting
 
 If `cmake` fails with...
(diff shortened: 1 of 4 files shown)


KitsuneRal:
Thanks, that's quite useful. I'm ok to have org.kde.Sdk for a build platform but I'd prefer to have some more compact runtime than org.kde.Platform for those GNOME and, even more, LXDE guys.

elvisangelaccio:
I tried to use the runtime from Liri OS, which ships only Qt5, but it doesn't work (diff attached if anyone wants to give it a shot).

[liri-runtime-patch.txt](https://github.com/Fxrh/Quaternion/files/973704/liri-runtime-patch.txt)

For the record, I don't think that the size of org.kde.Platform is an issue. The whole point of flatpak runtimes imho is to have a single collection of dependencies that can be easily re-used across many applications. For example, Qupzilla is a Qt-only browser that is also using the kde runtime for their flatpak package.

KitsuneRal:
...and therefore requires KF5 be installed even if I'm trying to setup a lean LXDE environment. Truth be told, I don't know how much this platform takes compared to an imaginary Qt5-only platform but that basically closes the door for Flatpak if/when I try to do anything compact. &lt;span class="rant"&gt;I guess it's a sign of the time though - people don't care to have 3 Docker containers on one machine that duplicate each other's environment almost entirely... And Snap/Flatpak don't make the situation much better, at least as of yet.lt;/span&gt;

KitsuneRal:
Anyway, I think I can take this in, given that we have enough KDE folks around to make use of it. I am personally not going to use it, though; I have a mostly GNOME environment with a few Qt-based pieces, and KDE libraries would just consume space without any benefit for me.

elvisangelaccio:
Thanks for merging :) 
Yes, flatpak is mainly meant for the average user that just wants an easy way to install an app, imho. If you are a power user that wants to minimize disk space, flatpak may not be desiderable.

KitsuneRal:
Unfortunately, Flatpak is convenient for a very limited set of average users. Just today stumbled over this, accidentally: flatpak/flatpak#114 - basically, this brings applications integration back to the times before fd.o. I still think that Snap and Flatpak are only good for single-app, or very-limited-app-set cases (i.e. kiosks), and most average users easily fall out of this.
Ok, I'm done with this, not gonna comment further.

PR #166, opened 2017-06-29
Fix flatpak startup with nvidia drivers

elvisangelaccio:
We need this argument in order to start the Quaternion flatpak with
nvidia drivers, otherwise we can't create the openGL context.


KitsuneRal:
Since I don't have anything like that around (only Intel), I have to just accept it as is :) Thanks!

PR #174, opened 2017-08-22
Don't cut the slider in the chat view

elvisangelaccio:
This adds a right margin to the chat view so that the Slider doesn't get
cut. We also need to adjust the width of a bunch of other elements to
avoid regressions (such as Slider overlapping the ToolButtons).

### DIFF
--- a/client/chatroomwidget.cpp
+++ b/client/chatroomwidget.cpp
@@ -19,6 +19,8 @@
 
 #include "chatroomwidget.h"
 
+#include <QtWidgets/QApplication>
+#include <QtWidgets/QStyle>
 #include <QtWidgets/QVBoxLayout>
 #include <QtWidgets/QLabel>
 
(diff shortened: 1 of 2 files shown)


elvisangelaccio:
Before:

before

After:

after

KitsuneRal:
Thanks, but I'm a bit wary of a "magic number" 10. Could you rather bind it to the slider width instead?

In fact, the original way I made it was due to a rather default width of a slider on Windows. So I just slashed it in half by `anchors.horizontalCenter: chatView.right`. So the right way of fixing this issue, probably, would include setting the slider width explicitly.

elvisangelaccio:
After the new patch:

spectacle c12100

KitsuneRal:
Here's how it look on Windows now - kind of ugly :-|
capture
Hardwiring the width to something about 10 (still a "magic number"...) makes it look better:
capture
I propose, if it doesn't break the look on Linux, to stick with Slider.width: 10, as of now. Maybe some later day somebody finally gets to make a custom widget for that.

elvisangelaccio:
@KitsuneRal A width of 10 breaks all the non-windows styles, so I added a style check.

KitsuneRal:
Now it's ugly in the code %-)
Ok, I'm going to take the first commit in, anyway. Meanwhile I experimented a bit with custom slider styles and I think I'll apply something like below (universally across platforms, using theme colours), while there's no better design. That thing behaves not as a usual slider so it's not really supposed to look like one.
capture

PR #175, opened 2017-09-02
change "runtime-version" from master to 5.9

alxlg:
During the Flatpak building it can't find org.kde.Sdk master version, instead version 5.9 is the only available currently. Changing it in the .json file make the Flatpak package build correctly.


elvisangelaccio:
+1, thanks.

PR #177, opened 2017-09-30
Linkify URLs enclosed in angle brackets

ryco117:
Attempt fix for issue https://github.com/QMatrixClient/Quaternion/issues/171

### DIFF
--- a/client/quaternionroom.cpp
+++ b/client/quaternionroom.cpp
@@ -106,7 +106,7 @@ void QuaternionRoom::linkifyUrls(QString& text) const
 // starting with a letter (not a digit).
 #define FQDN "(\\w[-\\w]*\\.)+(?!\\d)\\w[-\\w]+"
 // https://stackoverflow.com/a/7109208
-#define VALID_URI "[A-Za-z0-9%\\._~:\\/?#\\[\\]@!$&'\\(\\)*+,\\;=-]*"
+#define VALID_URI "((?!&gt;)[A-Za-z0-9%\\._~:\\/?#\\[\\]@!$&'\\(\\)*+,\\;=-])*"
     static const QRegularExpression urlDetectorMail {
         QStringLiteral("(^|\\s)(" // Criteria to match the beginning of the URL
             "\\w[^@/#\\s]*@" // authentication (the part before @)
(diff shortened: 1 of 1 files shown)


KitsuneRal:
Thanks for the PR. What if the URL is only prepended by a left bracket but doesn't have a trailing `>`; or conversely, what if there's no `<` but there is `>`?
I wonder if a lookahead for `&gt;` would help you to filter out the trailing `>`.

ryco117:
Corrected some comments and valid URI regex. Used lookahead to enforce that it cant end with "&gt;" if it doesn't start with '<' (ie nothing is linkified now if only one bracket)

screenshot from 2017-09-30 22-05-20

KitsuneRal:
Sorry for taking time, the week is very packed... I'm puzzled by your choice to not linkify URLs if the surrounding brackets are unbalanced. Do you have a certain reason for that? I actually should have been more explicit in my original comment - my take is that URLs should be linkified regardless of whether `<` and `>` exist around them. Or am I missing something?

ryco117:
Originally I went out of my way to do it like that because thats what I thought you wanted (you know what they say about when you assume..) but I'm now seeing that allowing it to linkify when it has an unmatched closing bracket (ie "http://test.test/test>" ) will result in an issue in HTML (=> "http://test.test/test&gt;") where the regex parser will have no indication that the "&gt;" is not part of the URI (so it'll linkify into the wrong address)

KitsuneRal:
Uhm, shouldn't `>` be percent-escaped if it occurs in a URL?

Interestingly, the github linkifier only includes the literal `>`, not `&gt;`. Doesn't look neither proper, nor consistent anyway.

KitsuneRal:
Oh, I start getting it. Since HTML-escaping comes first, the trailing `>` is converted to `&gt;`, which in turn is incorrectly parsed as a part of a URL by the linkifier :-\

KitsuneRal:
...so that means that the regex should treat the trailing `&gt;` as an HTML-escaped `>` and not include it in the URL. I guess it's better to move to the Matrix room with all that.

ryco117:
screenshot from 2017-10-05 22-32-38

Simpler code/regex, more desirable result

KitsuneRal:
Yes - almost perfect! But I think that `\\b` should be enough _before the URL_ is superfluous because `\\s` before `\\w` implies `\\b`. Or am I again missing something?

PR #205, opened 2017-11-17
Introduce AppStream metadata

elvisangelaccio:
AppStream specification: https://www.freedesktop.org/software/appstream/docs/

The current metadata file is very minimal, but it passes `appstreamcli validate`


KitsuneRal:
What about adding a screenshot? As the saying goes, programs that have screenshots get installed considerably more often :)

KitsuneRal:
Thanks a lot!

PR #230, opened 2017-12-18
Borrow Konsole's regexp for linkifyUrls method

### LINKED ISSUE #224: Linkifier still skips on some URLs
Some very obvious URLs like https://www.google.com are not linkified in Quaternion, even when sent from Quaternion (i.e. in plain text).

marcdeop:
Closes #224

### DIFF
--- a/client/quaternionroom.cpp
+++ b/client/quaternionroom.cpp
@@ -108,45 +108,26 @@ void linkifyUrls(QString& htmlEscapedText)
 {
     static const auto RegExpOptions =
         QRegularExpression::CaseInsensitiveOption
-#if (QT_VERSION >= QT_VERSION_CHECK(5, 4, 0))
-        | QRegularExpression::OptimizeOnFirstUsageOption
-#endif
+        #if (QT_VERSION >= QT_VERSION_CHECK(5, 4, 0))
+            | QRegularExpression::OptimizeOnFirstUsageOption
+        #endif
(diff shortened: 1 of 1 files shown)


KitsuneRal:
Please bring back that `#ifdef` around `QRegularExpression::OptimizeOnFirstUsageOption` - it's only available in Qt 5.4 and later, and Quaternion 0.0.5 is still supposed to compile with Qt as ancient as 5.2.1. This is why Travis CI fails at your PR.

KitsuneRal:
Well, except that this code does fix #224, it brings a few degradations:
- `mailto` is not linkified in `mailto:links@like.that`. Should be easily fixed though.
- It linkifies a trailing angle bracket (`>`) though angle brackets should appear unescaped in URLs.
Other than that, things look really nice; the regexes are smaller, IDN-friendly and don't linkify what shouldn't be.

marcdeop:
I fixed the `mailto:` link.

However... I can't seem to figure out **why** at `>` would be linked. I have spend quite some time on this and everywhere I have tried the regexp it says that it is *not*! ( test done [here for example](https://regexr.com/))

KitsuneRal:
That's indeed very puzzling: the RE just doesn't leave options to go through angle brackets after the scheme/leading www.
I think I'll accept this for now, and workaround that angle bracket somehow separately. Thanks a lot for the contribution!

PR #267, opened 2018-01-14
Minimize/restore by clicking the tray icon

marcdeop:


client/mainwindow.cpp
```diff
@@ -666,3 +667,32 @@ void MainWindow::closeEvent(QCloseEvent* event)
...
 }
 
+void MainWindow::systemTrayIconAction(QSystemTrayIcon::ActivationReason reason)
+{
+    switch (reason)
+    {
+        case QSystemTrayIcon::Unknown:
+        case QSystemTrayIcon::Context:
+        case QSystemTrayIcon::DoubleClick:
+        case QSystemTrayIcon::Trigger:
```
KitsuneRal: I would agree with people in #quaternion - it's better if it only show/hides at `Trigger` and `DoubleClick`. Other actions can be left unattended for now.

client/mainwindow.cpp
```diff
@@ -82,6 +82,7 @@ MainWindow::MainWindow()
     createMenu();
     systemTray = new SystemTray(this);
     systemTray->show();
+    connect( systemTray, SIGNAL(activated(QSystemTrayIcon::ActivationReason)), this, SLOT(systemTrayIconAction(QSystemTrayIcon::ActivationReason)) );
```
KitsuneRal: Unless you need dynamic binding or overloaded signals, try to use the static version of connect:
`connect( systemTray, &SystemTray::activated, this, &MainWindow::systemTrayIconAction);`
The difference is that the static form will check its validity (spelling of signals, compatibility of arguments etc.) at compile-time, while the form you used checks the validity at runtime.

client/mainwindow.cpp
```diff
@@ -666,3 +667,32 @@ void MainWindow::closeEvent(QCloseEvent* event)
...
+    {
+        hide();
+    }
+    else
+    {
+        show();
+        raise();
+        setFocus();
+    }
+}
```
KitsuneRal: I'd probably do the whole thing inside `SystemTray` - it's quite a compact class so far (unlike `MainWindow`), and what you need to do very much fits the `QWidget` (not even `QWindow`) interface that `SystemTray` now uses for `m_parent`. Can you arrange that?

marcdeop:
@KitsuneRal I re-implemented the functionality following your suggestions. May you have a second look please?

PS: I also renamed the systemtray class to systemtrayicon

client/systemtrayicon.cpp
```diff
@@ -45,3 +46,34 @@ void SystemTrayIcon::highlightCountChanged(QMatrixClient::Room* room)
...
+
+void SystemTrayIcon::systemTrayIconAction(QSystemTrayIcon::ActivationReason reason)
+{
+    switch (reason)
+    {
+        case QSystemTrayIcon::Unknown:
+        case QSystemTrayIcon::Context:
+        case QSystemTrayIcon::DoubleClick:
+        case QSystemTrayIcon::MiddleClick:
+            break;
```
KitsuneRal: You don't really need to list all the options - that's why `default` is there. I understand your desire to make the available options more prominent but that's really what Qt documentation is for (they may add more things, like `Wheel`, e.g., in later versions.
Overall, I'd edit this code as follows:
``` c++
switch (reason)
{
    case QSystemTrayIcon::Trigger:
    case QSystemTrayIcon::DoubleClick:
        showHide();
        break;
    default:
        ;
}
```

PR #271, opened 2018-01-16
feat: add Room Name to room tooltip in room list

marcdeop:
This adds "Room name" to the tool tip shown on the room list.

End result:

screen shot 2018-01-16 at 17 55 15


marcdeop:
How about now?

quaternion_room_tooltip

KitsuneRal:
Perfect, thank you!

PR #331, opened 2018-03-18
feat: add confirmation box before forgetting a room

### LINKED ISSUE #236: Forgetting should be confirmed by the user
Since this action is irreversible, Quaternion should ask the user to confirm the intention.

marcdeop:
Fixes #236

### DIFF
--- a/client/roomlistdock.cpp
+++ b/client/roomlistdock.cpp
@@ -22,6 +22,7 @@
 #include <QtWidgets/QMenu>
 #include <QtWidgets/QStyledItemDelegate>
 #include <QtWidgets/QInputDialog>
+#include <QtWidgets/QMessageBox>
 
 #include "models/roomlistmodel.h"
 #include "quaternionroom.h"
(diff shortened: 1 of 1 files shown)


client/roomlistdock.cpp
```diff
@@ -181,8 +182,16 @@ void RoomListDock::menuLeaveSelected()
 
 void RoomListDock::menuForgetSelected()
 {
-    if (auto room = getSelectedRoom())
-        room->connection()->forgetRoom(room->id());
+    if (auto room = getSelectedRoom()) {
+      QMessageBox confirmBox(QMessageBox::Question,
+          tr("Are you sure?"),
```
KitsuneRal: I'd rather use something like "Really forget the room?".

client/roomlistdock.cpp
```diff
@@ -181,8 +182,16 @@ void RoomListDock::menuLeaveSelected()
...
+          tr("Do you want to forget the room %1").arg(room->displayName()),
+          QMessageBox::Yes|QMessageBox::No, this);
+      if (confirmBox.exec() == QMessageBox::No)
+          ;
```
KitsuneRal: Just use comparison with `QMessageBox::Yes` and omit the else branch entirely instead?

client/roomlistdock.cpp
```diff
@@ -181,8 +182,16 @@ void RoomListDock::menuLeaveSelected()
 
 void RoomListDock::menuForgetSelected()
 {
-    if (auto room = getSelectedRoom())
-        room->connection()->forgetRoom(room->id());
+    if (auto room = getSelectedRoom()) {
+      QMessageBox confirmBox(QMessageBox::Question,
+          tr("Are you sure?"),
+          tr("Do you want to forget the room %1").arg(room->displayName()),
```
KitsuneRal: I'd rather force a "disambiguated" name here (meaning - attach canonical alias even in the case when the name exists, and even add a room id if neither room name nor canonical alias exist). The latter one is dubious because people don't know room id's but most people certainly know canonical aliases.

KitsuneRal (changes requested):

PR #361, opened 2018-06-17
Fix AppStream info

rushsteve1:
There were issues with the AppStream file that caused Flatpak builds to fail. This fixes the issue (had to alter the screenshot too) and updated the flatpak manifest file, and the .gitignore.


rushsteve1:
The Travis Build failed with something related to APT.
I have minimal experience with that and it is unrelated to my changes so I don't want to mess with it.

KitsuneRal:
I restarted Travis jobs and they went fine - there apparently was a glitch on Travis side.
I'm not quite getting the reason you updated the screenshot file and put it to an external resource. Can you please elaborate?
As for the AppData manifest in the source tree: this one (as opposed to the one at Flathub) is supposed to use the master branch by default. If it's a default for Flatpak (I don't know) then I'm fine to drop the `branch: master` line (the `tags` line looks not much relevant to me so let it go).

linux/com.github.quaternion.appdata.xml
```diff
@@ -21,7 +23,7 @@
 
   <screenshots>
     <screenshot type="default">
-      <image width="1300" height="740">https://raw.githubusercontent.com/QMatrixClient/Quaternion/master/quaternion.png</image>
+	    <image type="source" width="1294" height="734">https://my.mixtape.moe/ixwmja.png</image>
```
KitsuneRal: That doesn't seem to be necessary, except maybe adding `type="source"`.

rushsteve1: You are correct. That was for testing and I left it by accident.

rushsteve1:
To address your comment about pulling from the `master` branch I actually now have it set to build straight out of the development folder. So you can test your changes live in a Flatpak whenever just by rebuilding and reinstalling.

KitsuneRal:
Yes, I noticed - and I agree this should be the best way. Thanks, this one goes in!

PR #374, opened 2018-06-23
Added Direct Chat to menu

rushsteve1:
Added an option to the menu `Room`>`Direct Chat...` which opens a direct chat room easily with someone who you do not share a room with.

Most of this code was taken from the `joinRoom()` function.

### DIFF
--- a/client/mainwindow.cpp
+++ b/client/mainwindow.cpp
@@ -32,6 +32,7 @@
 #include "lib/networkaccessmanager.h"
 #include "lib/settings.h"
 #include "lib/logging.h"
+#include "lib/user.h"
 
 #include <QtCore/QTimer>
 #include <QtCore/QDebug>
@@ -158,6 +159,7 @@ void MainWindow::createMenu()
             summon(dlg, connections, this);
         });
     createRoomAction->setDisabled(true);
+    roomMenu->addAction(tr("&Direct chat..."), [=]{ directChat(); });
     roomMenu->addAction(tr("&Join room..."), [=]{ joinRoom(); } );
     roomMenu->addSeparator();
     roomMenu->addAction(tr("&Close current room"),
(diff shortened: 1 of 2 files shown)


client/mainwindow.cpp
```diff
@@ -735,6 +737,74 @@ void MainWindow::joinRoom(const QString& roomAlias, Connection* connection)
...
+            userName = userInput.textValue();
+        else
+            QMessageBox::warning(this, tr("No user id specified"),
+                                 tr("Please specify non-empty user id"),
+                                 QMessageBox::Close, QMessageBox::Close);
+    }
+
+    using QMatrixClient::BaseJob;
+    auto* user = new QMatrixClient::User(userName, connection);
+    user->requestDirectChat();
```
KitsuneRal: You're assuming here that this user exists on the server - that's a pretty strong assumption. Consder using `Connection::requestDirectChat` instead.

client/mainwindow.cpp
```diff
@@ -735,6 +737,74 @@ void MainWindow::joinRoom(const QString& roomAlias, Connection* connection)
...
+    statusBar()->showMessage(tr("Started chat with %1 as %2")
+                                .arg(userId, connection->userId()));
```
KitsuneRal: Opening the chat does not happen synchronously inside `requestDirectChat()` (no matter which one you call). You should connect to `directChatAvailable()` or just replace "Started" with "Starting" because once the chat is actually ready, Quaternion will switch to it, therefore providing very direct feedback to the user and there's no need to duplicate it in the status line.

client/mainwindow.cpp
```diff
@@ -735,6 +737,74 @@ void MainWindow::joinRoom(const QString& roomAlias, Connection* connection)
...
+        // TODO: Check validity, not only non-emptyness
+        if (!userInput.textValue().isEmpty())
+            userName = userInput.textValue();
+        else
+            QMessageBox::warning(this, tr("No user id specified"),
+                                 tr("Please specify non-empty user id"),
+                                 QMessageBox::Close, QMessageBox::Close);
+    }
+
+    using QMatrixClient::BaseJob;
```
KitsuneRal: You don't need that.

rushsteve1:
For now (with this PR anyway) I am going to hold off on the rewrite. Will probably do it next to be honest, but I just want to get the functionality landed first.

Issue #239, opened 2017-12-22
Dynamic linking of Quaternion with libqmatrixclient


KitsuneRal:
We're gradually coming to the state when libqmatrixclient has not only API but also ABI guarantees. Once this is done, it should be possible to build Quaternion separately from libqmatrixclient and link the two at runtime. This will please Linux distributions (they are rather unhappy when a library cannot be packaged separately); and who knows, maybe we'll start getting other clients and bindings in other languages thanks to that.

awilfox:
> The library is still built statically into Quaternion; if you'd like to take part in uncoupling the two, you're welcome in #239.

How can I help?  I'm the project lead of [one of those aforementioned Linux distributions](https://adelielinux.org) that would love to have the library decoupled; I'm a moderately skilled Qt programmer; and I am also a regular user of Quaternion.

awilfox:
Okay, so packaging Quaternion for Adélie instead of just running it from the build directory:

* It defaults to building a .so instead of an .a, at least here (GCC 6.4.0, binutils 2.29, ppc64 and x86_64)
* The .so file is not installed to /usr/lib, so the software fails to run due to a missing library:

```
awilcox on reynard [pts/6 Sat 30 16:25] ~: quaternion
Error loading shared library libqmatrixclient.so.0: No such file or directory (needed by /usr/bin/quaternion)
Error relocating /usr/bin/quaternion: _ZN13QMatrixClient4RoomD2Ev: symbol not found
```

After applying a [small patch](https://code.foxkit.us/adelie/packages/blob/master/user/quaternion/install-library.patch) to install the library, it seems to work fine.  If this patch would be helpful I would be happy to put it in a pull request.

KitsuneRal:
Thanks a lot! As far as I understand it, there are two parts here:
1. Make sure the library gets installed to the right location so that it can be packaged. Your patch does that - please submit it to the `QMatrixClient/libqmatrixclient` repo.
2. Make sure that versioning of the library is correct (not only .so but also .so.* files are generated). As far as I see above, that seems to work as well.

KitsuneRal:
Oh, sorry, there's one more thing - but that's really above compiling against the dynamically linked library. The library headers should be installed to the standard place (`/usr/include` or whatever the distribution uses). So while you're at it, I'd appreciate if you add one more line to your patch that takes all `.h` files from `lib/` and installs them.

KitsuneRal:
[...]

PR #383, opened 2018-07-13
Remove references to lib in all #include headers.

arunisaac:
This pull request removes references to lib in all #include headers, as discussed in #239.


KitsuneRal:
Thank you!

PR #401, opened 2018-08-20
Omitting OARS tags with the value of none

har9862:
OARS now supports a compressed tag listing, if the value is `none` it's not longer necessary to explicitly list a tag. This makes the list more readable and reduces the need for updating the file every time there is a change to the OARS specs.


KitsuneRal:
That's sweet. Is it still considered OARS 1.0 or should the version be updated too?

KitsuneRal:
As https://hughsie.github.io/oars/generate.html now makes the concise listing even for OARS 1.0, I assume it was a compatible change in the protocol. Thanks!

PR #414, opened 2018-09-24
Add icons in menus

kRkk:
before:
before

after:
after

### DIFF
--- a/client/mainwindow.cpp
+++ b/client/mainwindow.cpp
@@ -150,7 +150,7 @@ void MainWindow::createMenu()
     // Connection menu
     connectionMenu = menuBar()->addMenu(tr("&Accounts"));
 
-    connectionMenu->addAction(QIcon::fromTheme("contact-new"), tr("&Login..."),
+    connectionMenu->addAction(QIcon::fromTheme("im-user"), tr("&Login..."),
         this, [=]{ showLoginWindow(); } );
 
     connectionMenu->addSeparator();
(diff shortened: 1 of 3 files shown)


client/mainwindow.cpp
```diff
@@ -207,24 +208,29 @@ void MainWindow::createMenu()
     roomSettingsAction->setDisabled(true);
     roomMenu->addSeparator();
     createRoomAction =
-        roomMenu->addAction(tr("Create &new room..."), [this]
+        roomMenu->addAction(QIcon::fromTheme("mail-message-new"),
+        tr("Create &new room..."), [this]
```
KitsuneRal: I'm not quite sure about this one, to be frank. An envelope really doesn't convey the right idea. I suppose `document-new` is the closest metaphor.

client/mainwindow.cpp
```diff
@@ -207,24 +208,29 @@ void MainWindow::createMenu()
     roomSettingsAction->setDisabled(true);
```
KitsuneRal: While you're at it, adding an icon to `roomSettingsAction` (I propose `document-properties`) would be great.

client/mainwindow.cpp
```diff
@@ -150,14 +150,15 @@ void MainWindow::createMenu()
     // Connection menu
     connectionMenu = menuBar()->addMenu(tr("&Accounts"));
 
-    connectionMenu->addAction(tr("&Login..."), this, [=]{ showLoginWindow(); } );
+    connectionMenu->addAction(QIcon::fromTheme("contact-new"), tr("&Login..."),
+        this, [=]{ showLoginWindow(); } );
```
KitsuneRal: `network-server` maybe? I know it's not the action icon, according to the spec, but using the same icon for logging in and opening a direct chat looks really confusing.

client/mainwindow.cpp
```diff
@@ -305,7 +311,8 @@ void MainWindow::createMenu()
     );
 
     settingsMenu->addSeparator();
-    settingsMenu->addAction(tr("Configure &network proxy..."), [this]
+    settingsMenu->addAction(QIcon::fromTheme("network-wired"),
+        tr("Configure &network proxy..."), [this]
```
KitsuneRal: This should rather be `system-preferences-network`. Other network configuration options will be in the same dialog box as the network proxy configuration.

kRkk:
Menu:
accounts menu
View menu:
view
Room menu:
room
Settings menu:
settings
Room list dock:
room list

kRkk:
You can merge this

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

PR #418, opened 2018-10-04
WIP: Add translations

### LINKED ISSUE #394: Enable i18n
#295 formally enabled localisation but there's still no place to point translators to in order to make Quaternion available in other languages.

kRkk:
Closes #394

Tested only on Linux

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,7 +37,7 @@ foreach (FLAG "" all pedantic extra no-unused-parameter)
 endforeach ()
 
 # Find the libraries
-find_package(Qt5 5.6 REQUIRED Widgets Network Quick Qml QuickWidgets Gui)
+find_package(Qt5 5.6 REQUIRED Widgets Network Quick Qml QuickWidgets Gui LinguistTools)
 if (USE_QQUICKWIDGET)
     find_package(Qt5 5.6 REQUIRED QuickWidgets)
 endif()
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -108,9 +108,16 @@ set(quaternion_QRC
     client/resources.qrc
     )
 
+set(quaternion_TS
+    client/translations/quaternion.en.ts
+    )
```
KitsuneRal: I've just found a disappointing bug in Qt: https://bugreports.qt.io/browse/QTBUG-64317 - since we're still on Qt 5.6, it's probably better to use `quaternion_en.ts` instead.

CMakeLists.txt
```diff
@@ -108,9 +108,16 @@ set(quaternion_QRC
...
+
 QT5_ADD_RESOURCES(quaternion_QRC_SRC ${quaternion_QRC})
 set_property(SOURCE qrc_resources.cpp PROPERTY SKIP_AUTOMOC ON)
 
+QT5_ADD_TRANSLATION(quaternion_QM ${quaternion_TS})
+QT5_CREATE_TRANSLATION(client/translations ${quaternion_SRCS} ${quaternion_QRC_SRC} ${quaternion_TS})
```
KitsuneRal: Here's another bummer: https://bugreports.qt.io/browse/QTBUG-41736 - basically, `make clean` will kill your existing translations, remaking .ts files from the source code instead. Still unfixed, but the comments provide a workaround that looks good to me. Addressing this comment is not mandatory before merge.

client/main.cpp
```diff
@@ -51,8 +51,15 @@ int main( int argc, char* argv[] )
...
+    QString translationsDir;
+#if defined(Q_OS_WIN32)
+    translationsDir = "./translations";
+#else
+    translationsDir = "/usr/local/share/quaternion/translations";
+#endif
+
     QTranslator appTranslator;
-    appTranslator.load(QLocale(), "quaternion", ".", "i18n");
+    appTranslator.load(QLocale(), "quaternion", ".", translationsDir);
```
KitsuneRal: First off, it would be great to still support the case of running Quaternion directly from the build directory. Second, I'd rather try finding `translations/` subdirectory in all directories from `QStandardPaths::AppLocalDataLocation` and then call `load()` on the found one (if any). So I guess it would look like below:
```c++
if (!appTranslator.load(QLocale(), "quaternion", "_" /* instead of ".", see the comment above */))
    appTranslator.load(QLocale(), "quaternion", "_", QStandardPaths::locate(QStandardPaths::AppLocalDataLocation, "translations"));
```
but I haven't tested it.

CMakeLists.txt
KitsuneRal: One more thing: you have forgotten to pass `quaternion_QM` as the first parameter to `QT5_CREATE_TRANSLATION`, haven't you?

KitsuneRal:
Great, thanks!

PR #420, opened 2018-10-07
Add German translation

krombel:
This as well updates the source (the english file) to reflect the now current sources

There are some missing strings which are not translated yet but with a second translation they can be observed more easily


KitsuneRal:
Thanks!

PR #426, opened 2018-10-12
further german translations

krombel:


client/translations/quaternion_de_DE.ts
```diff
@@ -194,60 +194,61 @@
...
         <source>Cancel downloading</source>
-        <translation type="unfinished"></translation>
+        <translation>Herunterladen abbrechen</translation>
     </message>
     <message>
         <location filename="../qml/ImageContent.qml" line="59"/>
         <source>Open in image viewer</source>
-        <translation type="unfinished"></translation>
+        <translatorcomment>Is it OK to just state open externally?</translatorcomment>
+        <translation>In Bildbetrachter öffnen</translation>
```
krombel: It would be more intuitive to have this "Open externally" I think. Should this be changed?

KitsuneRal: Yes, probably.

KitsuneRal:
So be it :)

PR #438, opened 2018-10-24
Set inital device name to hostname

rushsteve1:
When adding an account when there are none set it to the device name.


KitsuneRal:
Thank you, but as discussed in the Matrix room, I'd really prefer it to be something less revealing. Maybe just give an example using `setPlaceholderText("e.g. Home Desktop")`.

KitsuneRal:
This has been accidentally merged (with the reverting commit immediately afterwards). If you still want to get this in, please cherry-pick the commit to another branch and make a new PR.

PR #442, opened 2018-10-25
displayname to displayName

rushsteve1:
The capitalization was inconsistent with other methods.

Goes with the matching branch on libqmatrixclient PR https://github.com/QMatrixClient/libqmatrixclient/pull/251.

This is waiting on the user searching PR #437 to be merged because it changes some things in that as well. So any merge issues at current are likely due to that.


KitsuneRal:
Not going to fix this for now: API compatibility is more important than spelling fixes. Maybe later along with some bigger change to display names API.

PR #443, opened 2018-10-28
Set the maximum height for ChatEdit

### LINKED ISSUE #246: Typing in a too long message completely hides the timeline
The message input is only limited by the caption area, disregarding the timeline. I guess it only should be allowed to grow halfway, and then rely on scrolling instead of further growing.

krkk:
Closes #246


KitsuneRal:
Sorry, will re-implement it on my end.

PR #456, opened 2018-12-16
fix macOS build instructions

ara4n:


BUILDING.md
```diff
@@ -61,13 +61,17 @@ On Fedora 26, the following command should be enough for building and running:
...
 ```
 
-#### OS X
-`brew install qt5` should get you Qt5. You may need to tell CMake about the path to Qt by passing `-DCMAKE_PREFIX_PATH=<where-Qt-installed>`
+#### macOS
+`brew install qt5` should get you Qt5. You have to point CMake at the qt5 installation location, with something like:
+
+```
+cmake .. -DCMAKE_PREFIX_PATH=../../libqmatrixclient -DCMAKE_PREFIX_PATH=$(brew --prefix qt5)
+```
```
KitsuneRal: Or maybe just propose to use `-DUSE_INTREE_LIBQMC=ON`? Unless the library is provided as a macOS framework of its own, I believe it cannot reused between two binaries anyway, can it?

ara4n: yup, you can't reuse it between two binaries, but i thought the reason to prefer an out-of-tree libqmatrixclient was to just manage it as an explicit dep rather than via git submodules. I've updated it to propose both now though.

KitsuneRal: To sum up on this: I will change the default to use the in-tree library shortly, and will update the building instructions accordingly. And for now I'm accepting this change as-is (because it reflects the reality as yet).

PR #457, opened 2018-12-19
Add icon to macOS build

aaronraimist:
I don't know anything about cmake but this seems to work, let me know if this looks right.


KitsuneRal:
LGTM - thanks!

PR #463, opened 2018-12-27
Set a few more macOS info.plist items

### LINKED ISSUE #462: broken Info.plist on macOS
most of the entries in your Info.plist file are empty, which hinders macOS native integration.

each app needs a bundle identifier (CFBundleIdentifier), not having it will break cache and preferences

each app needs version information (CFBundleVersion & CFShortVersionString), so you can see its version in the Finder

the CFBundleLongVersionString and CFBundleGetInfoString entries have been deprecated for > 10 years and can be removed

aaronraimist:
Fixes #462 

I used the same identifier as the linux packages `com.github.quaternion`. If there is a better thing to set that to, let me know.

I'm just setting CFBundleVersion and CFShortVersionString to the same thing, and I guess I'll just set the version manually. I wasn't sure if there was a way for cmake to "know" what the version is automatically.

Get Info screen looks like this after the PR:
screen shot 2018-12-27 at 12 42 18 am

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -152,11 +152,18 @@ if(WIN32)
 endif()
 
 if(APPLE)
+    set(MACOSX_BUNDLE_GUI_IDENTIFIER "com.github.quaternion")
+    set(MACOSX_BUNDLE_BUNDLE_NAME "quaternion")
+
+    set(MACOSX_BUNDLE_COPYRIGHT "Copyright © 2018 QMatrixClient Project")
+
+    set(VERSION "0.0.9.4")
+    set(MACOSX_BUNDLE_SHORT_VERSION_STRING ${VERSION})
+    set(MACOSX_BUNDLE_BUNDLE_VERSION ${VERSION})
+
     set(ICON_NAME "quaternion.icns")
     set(ICON_PATH "${CMAKE_CURRENT_SOURCE_DIR}/icons/${ICON_NAME}")
-
     set(MACOSX_BUNDLE_ICON_FILE ${ICON_NAME})
-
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -152,11 +152,18 @@ if(WIN32)
 endif()
 
 if(APPLE)
+    set(MACOSX_BUNDLE_GUI_IDENTIFIER "com.github.quaternion")
+    set(MACOSX_BUNDLE_BUNDLE_NAME "quaternion")
+
+    set(MACOSX_BUNDLE_COPYRIGHT "Copyright © 2018 QMatrixClient Project")
+
+    set(VERSION "0.0.9.4")
```
KitsuneRal: Can we please put this one `set` outside `if(APPLE)`, before `if(WIN32)`? I think I'll take it for some other good uses as well.

aaronraimist: I moved it up near the top of the file since there are two `if(WIN32)`s. Plus this will make it easy to find when a new version comes out.

CMakeLists.txt
```diff
@@ -152,11 +154,17 @@ if(WIN32)
 endif()
 
 if(APPLE)
+    set(MACOSX_BUNDLE_GUI_IDENTIFIER "com.github.quaternion")
+    set(MACOSX_BUNDLE_BUNDLE_NAME "quaternion")
+
+    set(MACOSX_BUNDLE_COPYRIGHT "Copyright © 2018 QMatrixClient Project")
```
KitsuneRal: Sorry, just noticed - can you please move this out and put next to `VERSION` as well (and make it `COPYRIGHT` or smth like that, of course)? Apparently that's one more hardcoded thing that's worth being put to CMakeLists.txt.

aaronraimist: 👍

PR #474, opened 2019-01-18
Macdeployqt set -qmldir and fix if not in path

### LINKED ISSUE #473: macdeployqt not found if not in PATH
A similar trick to the one with `windeployqt` (with `get_filename_component`) should be done to invoke macdeployqt.

aaronraimist:
Fixes #473

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -245,7 +245,7 @@ endif(WIN32)
 
 if(APPLE)
     set(MACDEPLOYQT ${Qt5Widgets_DIR}/../../../bin/macdeployqt)
-    set(MACDEPLOYQT_ARGS ${CMAKE_CURRENT_BINARY_DIR}/${APP_NAME}.app -dmg -qmldir=../client/qml)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -244,8 +244,8 @@ if(WIN32)
 endif(WIN32)
 
 if(APPLE)
-    set(MACDEPLOYQT macdeployqt)
-    set(MACDEPLOYQT_ARGS ${CMAKE_CURRENT_BINARY_DIR}/${APP_NAME}.app -dmg)
+    set(MACDEPLOYQT ${Qt5Widgets_DIR}/../../../bin/macdeployqt)
+    set(MACDEPLOYQT_ARGS ${APP_NAME}.app -dmg -qmldir=../client/qml)
```
KitsuneRal:
```suggestion
    set(MACDEPLOYQT_ARGS ${APP_NAME}.app -dmg -qmldir="${PROJECT_SOURCE_DIR}/client/qml")
```
because you cannot rely on the build directory always being one under the sources directory.

PR #477, opened 2019-01-30
call activateWindow() in addition to raise()/setFocus()

ddanilov:
for me (X11) rise() does not bring the window in front of all other windows in all cases, activateWindow() does

### DIFF
--- a/client/systemtrayicon.cpp
+++ b/client/systemtrayicon.cpp
@@ -80,6 +80,7 @@ void SystemTrayIcon::showHide()
     else
     {
         m_parent->show();
+        m_parent->activateWindow();
         m_parent->raise();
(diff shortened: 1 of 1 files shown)


KitsuneRal:
FTR: Upon reading documentation I start to wonder if `raise()` should even be there; the doc on `activateWindow()` says that the method is equivalent to clicking on the title bar of the window, usually (but not always, depending on your WM - and those using such WM setting had some reason for it, right?) leading to raising the window. On the other hand, `raise()` apparently doesn't even always work.

Also, later down the road I'd probably make `activateWindow()` invocation optional as well (when we actually have a preferences dialog), for folks that want to be able to show/hide Quaternion preserving its stack order amongst other windows.

PR #484, opened 2019-02-05
 add support for qtkeychain to store access tokens

ddanilov:


client/mainwindow.cpp
```diff
@@ -421,8 +434,72 @@ QByteArray MainWindow::loadAccessToken(const AccountSettings& account)
...
+#ifdef USE_KEYCHAIN
+QByteArray MainWindow::loadAccessTokenFromKeyChain(const QMatrixClient::AccountSettings &account)
+{
+    // check for existing token file
+    auto accessToken = loadAccessTokenFromFile(account);
+    if(!accessToken.isEmpty())
+    {
+        if(QMessageBox::warning(this,
+                                tr("Access token file found"),
+                                tr("Do you want to migrate the access token for ")
```
tsdgeos: This is a word puzzle that makes translations harder, should be tr("Bla Bla Bla %1 Bla Bla Bla").arg(account.userId())

ddanilov: ok, will change that

client/mainwindow.cpp
```diff
@@ -465,6 +540,36 @@ bool MainWindow::saveAccessToken(const AccountSettings& account,
     return false;
 }
 
+#ifdef USE_KEYCHAIN
+bool MainWindow::saveAccessTokenToKeyChain(const QMatrixClient::AccountSettings &account,
```
KitsuneRal:
```suggestion
bool MainWindow::saveAccessTokenToKeyChain(const AccountSettings& account,
```
`QMatrixClient::AccountSettings` is under `using` throughout the file; also aligning ampersand positioning with the rest of the codebase.

client/mainwindow.cpp
```diff
@@ -421,8 +434,70 @@ QByteArray MainWindow::loadAccessToken(const AccountSettings& account)
     return {};
 }
 
+#ifdef USE_KEYCHAIN
+QByteArray MainWindow::loadAccessTokenFromKeyChain(const QMatrixClient::AccountSettings &account)
```
KitsuneRal:
```suggestion
QByteArray MainWindow::loadAccessTokenFromKeyChain(const AccountSettings& account)
```

client/mainwindow.cpp
```diff
@@ -465,6 +540,36 @@ bool MainWindow::saveAccessToken(const AccountSettings& account,
...
+{
+    qDebug() << "Save access token to keychain for " << account.userId();
+    QKeychain::WritePasswordJob job(qAppName());
+    job.setAutoDelete(false);
+    job.setKey(account.userId());
+    job.setBinaryData(accessToken);
+    QEventLoop loop;
+    QKeychain::WritePasswordJob::connect(&job, &QKeychain::Job::finished, &loop, &QEventLoop::quit);
+    job.start();
+    loop.exec();
```
KitsuneRal: I understand why the access token should be loaded synchronously (although even that could probably be aligned with the login flow, which is asynchronous - but that's more work so not for now probably); but saving the token can certainly be done asynchronously, can't it?

client/mainwindow.cpp
```diff
@@ -788,6 +893,24 @@ void MainWindow::logout(Connection* c)
...
     QFile(accessTokenFileName(AccountSettings(c->userId()))).remove();
 
+#ifdef USE_KEYCHAIN
+    QKeychain::DeletePasswordJob job(qAppName());
+    job.setAutoDelete(false);
+    job.setKey(c->userId());
+    QEventLoop loop;
+    QKeychain::DeletePasswordJob::connect(&job, &QKeychain::Job::finished, &loop, &QEventLoop::quit);
+    job.start();
+    loop.exec();
```
KitsuneRal: Same as for saving - probably this can be done asynchronously (by starting the job and letting it finish within the main event loop). Logging out is likely to take longer time anyway, these can be done in parallel.

client/mainwindow.cpp
```diff
@@ -465,6 +540,36 @@ bool MainWindow::saveAccessToken(const AccountSettings& account,
...
+
+    if (job.error())
+    {
+        qWarning() << "Could not save access token to keychain: " << qPrintable(job.errorString());
+        QMessageBox::warning(this,
+                             tr("Couldn't save access token"),
+                             tr("Quaternion couldn't save the access token to keychain."
+                                " You're logged in but will have to provide your password"
+                                " again when you restart the application."),
+                             QMessageBox::Close);
```
KitsuneRal: Can you provide a fallback to legacy files if QtKeychain is there but a backend could not be found? Too bad that (it seems after reading QtKeychain code) it cannot be detected upfront but at least we could recover from the error afterwards?

client/mainwindow.cpp
ddanilov: > ...consider adding QtKeychain to `.travis.yml` 

I'll try it: ubuntu xenial has a version of qt5keychain-dev and homebrew seems to mention qtkeychain as a package as well

> I understand why the access token should be loaded synchronously (although even that could probably be aligned with the login flow, which is asynchronous - but that's more work so not for now probably); but saving the token can certainly be done asynchronously, can't it?

sure it can be done asynchronously. But then the question

> Can you provide a fallback to legacy files if QtKeychain is there but a backend could not be found?

comes across... The current implementation mimics the synchronous way of the existing save/load/delete logic of storing the token in the file, it is sort of drop-in replacement. To make the keychain interaction  asynchronous, I will need to either decouple it from the file handling logic completely or refactor the handling logic to be asynchronous (with all the signals and slots) as well.

KitsuneRal: Ok, since this is most likely a file operation in the end (even if masqueraded by an "backend API call"), let's keep it as is for now, and see if this can/should be rearranged later.

BUILDING.md
```diff
@@ -51,12 +51,16 @@ hacking on Quaternion _and_ libQMatrixClient at the same time.
...
+- optionaly [QtKeychain](https://github.com/frankosterfeld/qtkeychain) to store access tokens in libsecret keyring or similar providers.
 
 #### Linux
 Just install things from the list above using your preferred package manager. If your Qt package base is fine-grained you might want to take a look at `CMakeLists.txt` to figure out which specific libraries Quaternion uses (or blindly run cmake and look at error messages). Note also that you'll need several Qt Quick plugins for Quaternion to work (without them, it will compile and run but won't show the messages timeline). In case of Xenial Xerus following line should get you everything necessary to build and run Quaternion:
 ```bash
 sudo apt-get install git cmake qtdeclarative5-dev qtdeclarative5-qtquick2-plugin qtdeclarative5-controls-plugin qml-module-qtquick-controls qml-module-qtquick-controls2 qtmultimedia5-dev
 ```
+```bash
+sudo apt-get install qt5keychain-dev
+```
```
KitsuneRal: Can you please add a one-line explanation like "if you want to use QtKeychain, add this" before this addition?

CMakeLists.txt
```diff
@@ -42,7 +42,7 @@ foreach (FLAG "" all pedantic extra no-unused-parameter)
 endforeach ()
 
 # Find the libraries
-find_package(Qt5 5.7 REQUIRED Widgets Network Quick Qml Gui LinguistTools Multimedia)
+find_package(Qt5 5.7 REQUIRED Widgets Network Quick Qml Gui LinguistTools Multimedia DBus)
```
KitsuneRal: DBus is needed only for Linux, no? Besides, it's a transitive dep of QtKeychain so probably keep it there?

client/mainwindow.cpp
ddanilov: done

client/mainwindow.cpp
ddanilov: done

client/mainwindow.cpp
ddanilov: it may have a user interaction in-between, for example when the keyring is locked and we wait for user to type-in a password to unlock it or so... but for now I also like to keep it in this form

client/mainwindow.cpp
ddanilov: done

BUILDING.md
ddanilov: done

CMakeLists.txt
ddanilov: that's still unclear for me: my local debian build is fine without it, the travis osx also. I have to add DBus there just to make the travis ubuntu build happy, maybe something is "wrong" with the ppa Qt pack? Could it break windosw builds?

KitsuneRal: On Windows Qt usually comes as one big bundle, and DBus is a part of it (even if not really operational).
Not sure what goes wrong here; QtKeychain seems to diligently add DBus deps to the public list of its libraries. Assuming that DBus (almost?) always comes together with QtCore, I think I'll just accept it for now.

aaronraimist:
> I'm fine to merge even if this only works on Linux.

Also seems to work great on Mac!

screen shot 2019-02-19 at 8 09 45 am

PR #488, opened 2019-02-09
Change user highlight to use RegEx

redsky17:
User highlighting is now based on Regular Expression instead of simple 'contains.'  This should help prevent errors in highlighting when it shouldn't be done.

Here's a picture showing that it still works :)

image

### DIFF
--- a/client/quaternionroom.cpp
+++ b/client/quaternionroom.cpp
@@ -21,7 +21,7 @@
 
 #include <user.h>
 #include <events/roommessageevent.h>
-#include <QRegularExpression>
+#include <QtCore/QRegularExpression>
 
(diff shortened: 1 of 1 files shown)


client/quaternionroom.cpp
```diff
@@ -21,6 +21,7 @@
 
 #include <user.h>
 #include <events/roommessageevent.h>
+#include <QRegularExpression>
```
KitsuneRal:
```suggestion
#include <QtCore/QRegularExpression>
```

KitsuneRal:
Thanks!

medusacle:
that was fast thank you !

PR #489, opened 2019-02-09
ImageContent: align image to the left

krkk:


KitsuneRal:
Yeah, probably, that would be better for small images (ehem, stickers inclusive).

PR #497, opened 2019-02-15
Timeline menu

### LINKED ISSUE #158: Context menu for messages
We tried it sometime ago, only to find that QML has compatibility issues in positioning the menu on the viewport. Since an ellipsis button (at least, per se) is not really scalable in terms of functionality, and there's already a demand for, e.g., "Copy permalink", someone needs to get to the bottom of it and figure out how to make it work.

One possible option would be to (a bit counter-intuitively, and not friendly to Tensor) generate a context menu on the C++ side, using Qt Widgets instead. Or make a custom context menu in QML - shouldn't be too hard.

### LINKED ISSUE #422: Link: Allow copy to clipboard instead of opening in browser
e.g. per right click. It is currently possible to select the whole text but that is inconvenient

krkk:
Closes #158 and closes #422
I don't know if some buttons in FileContent such as "Save as…" and "Open Folder" should remain.

### DIFF
--- a/client/chatroomwidget.cpp
+++ b/client/chatroomwidget.cpp
@@ -25,6 +25,9 @@
 #include <QtWidgets/QAction>
 #include <QtWidgets/QFileDialog>
 #include <QtWidgets/QApplication>
+#include <QtWidgets/QMenu>
+#include <QtGui/QClipboard>
+#include <QtGui/QDesktopServices>
 
 #include <QtQml/QQmlContext>
 #include <QtQml/QQmlEngine>
@@ -660,6 +663,59 @@ void ChatRoomWidget::quote(const QString& htmlText)
     }
 
     m_chatEdit->insertPlainText(sendString);
+}
+
+void ChatRoomWidget::showMenu(int index, bool showingDetails)
+{
+    const auto modelIndex = m_messageModel->index(index, 0);
+    const auto eventId = modelIndex.data(MessageEventModel::EventIdRole).toString();
+
+    QMenu menu;
+    menu.addAction(QIcon::fromTheme("edit-delete"), tr("Redact"), [=] {
+        m_currentRoom->redactEvent(eventId);
+    });
+    menu.addAction(QIcon::fromTheme("link"), tr("Copy permalink to clipboard"), [=] {
+        QApplication::clipboard()->setText("https://matrix.to/#/" +
+            m_currentRoom->id() + "/" + QUrl::toPercentEncoding(eventId));
+    });
+    menu.addAction(QIcon::fromTheme("format-text-blockquote"),
+                   tr("Quote", "a verb (do quote), not a noun (a quote)"), [=] {
+        emit quote(modelIndex.data().toString());
+    });
+    auto a = menu.addAction(QIcon::fromTheme("view-list-details"), tr("Show details"), [=] {
+        emit showDetails(index);
+    });
+    a->setCheckable(true);
+    a->setChecked(showingDetails);
+
+    const auto eventType = modelIndex.data(MessageEventModel::EventTypeRole).toString();
+    if (eventType == "image" || eventType == "file")
+    {
+        const auto progressInfo = modelIndex.data(MessageEventModel::SpecialMarksRole)
+            .value<QMatrixClient::FileTransferInfo>();
+        const bool downloaded = !progressInfo.isUpload && progressInfo.completed();
+
+        menu.addSeparator();
+        menu.addAction(QIcon::fromTheme("document-open"), tr("Open externally"), [=] {
+            emit openExternally(index);
+        });
+        menu.addAction(QIcon::fromTheme("folder-open"), tr("Open Folder"), [=] {
+            if (!downloaded)
+                m_currentRoom->downloadFile(eventId);
+
+            QDesktopServices::openUrl(progressInfo.localDir);
+        });
+        if (!downloaded)
+        {
+            menu.addAction(QIcon::fromTheme("edit-download"), tr("Download"), [=] {
+                m_currentRoom->downloadFile(eventId);
+            });
+        }
+        menu.addAction(QIcon::fromTheme("document-save-as"), tr("Save file as..."), [=] {
+            saveFileAs(eventId);
+        });
+    }
+    menu.exec(QCursor::pos());
 }
 
 void ChatRoomWidget::reStartShownTimer()
--- a/client/chatroomwidget.cpp
+++ b/client/chatroomwidget.cpp
@@ -665,7 +665,8 @@ void ChatRoomWidget::quote(const QString& htmlText)
     m_chatEdit->insertPlainText(sendString);
 }
 
-void ChatRoomWidget::showMenu(int index, bool showingDetails)
+void ChatRoomWidget::showMenu(int index, const QString& hoveredLink,
+                              bool showingDetails)
 {
     const auto modelIndex = m_messageModel->index(index, 0);
     const auto eventId = modelIndex.data(MessageEventModel::EventIdRole).toString();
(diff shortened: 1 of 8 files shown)


client/qml/TimelineItem.qml
```diff
@@ -389,29 +389,101 @@ Item {
                             Math.min(implicitHeight, textField.height) :
                             implicitHeight
 
-                text: "..."
+                text: "…"
+
+                // NOTE: in Qt < 5.12.0 using QQuickView
+                // the menu moves outside the button
+                // https://bugreports.qt.io/browse/QTBUG-40883
+                menu: Menu {
```
KitsuneRal: Can we make it popup on right click and dismiss the tool button altogether?

client/qml/TimelineItem.qml
```diff
@@ -389,29 +389,101 @@ Item {
...
+                            }
+                        }
+                    }
+
+                    MenuSeparator { }
+
+                    MenuItem {
+                        text: qsTr("Open externally")
+                        iconName: "document-open"
+                        enabled: eventType == "image" || eventType == "file"
```
KitsuneRal:
```suggestion
                        visible: eventType == "image" || eventType == "file"
```
For other events, this item doesn't even make sense, so no need to clutter the menu.

client/qml/TimelineItem.qml
```diff
@@ -389,29 +389,101 @@ Item {
...
+                        onTriggered: attachment.openExternally()
+                    }
+
+                    MenuItem {
+                        text: qsTr("Open Folder")
+                        iconName: "folder-open"
+                        enabled: (eventType == "image" || eventType == "file") &&
```
KitsuneRal:
```suggestion
                        visible: eventType == "image" || eventType == "file"
```

client/qml/TimelineItem.qml
```diff
@@ -389,29 +389,101 @@ Item {
...
+                                 progressInfo && progressInfo.localDir
```
KitsuneRal:
```suggestion
                        enabled: progressInfo && progressInfo.localDir
```

client/qml/TimelineItem.qml
```diff
@@ -389,29 +389,101 @@ Item {
...
+                                room.downloadFile(eventId)
+
+                            Qt.openUrlExternally(progressInfo.localDir)
+                        }
+                    }
+
+                    MenuItem {
+                        text: qsTr("Download")
+                        iconName: "edit-download"
+                        enabled: ((eventType == "image" && !attachment.autoload) ||
```
KitsuneRal: I guess you already get the idea from the suggestions above; please apply the same to this and the next item.

client/qml/TimelineItem.qml
```diff
@@ -389,29 +389,101 @@ Item {
...
+                        onTriggered: controller.saveFileAs(eventId)
                     }
-                    NumberAnimation {
-                        target: detailsAreaLoader; property: "opacity"
-                        to: showDetails.checked
-                        duration: settings.fast_animations_duration_ms
-                        easing.type: Easing.OutQuad
+
+
+                    TextEdit {
```
KitsuneRal:
```suggestion
                    TextField {
```
It was TextEdit in the original code but I really don't know what I was thinking about back then: TextField is much more lightweight. We also don't need to set `renderType`, since it renders to a zero-sized rectangle.

client/chatroomwidget.cpp
```diff
@@ -658,6 +661,70 @@ void ChatRoomWidget::quote(const QString& htmlText)
     m_chatEdit->insertPlainText(sendString);
 }
 
+void ChatRoomWidget::showMenu(const QString& hoveredLink, bool showingDetails,
+                              int index)
+{
+    const QString eventType = m_messageModel->data(m_messageModel->index(index, 0),
```
KitsuneRal:
```suggestion
    const auto eventType = m_messageModel->index(index, 0)->data(
```
And you really should put `m_messageModel->index(index, 0)` to a separate variable, instead of calculating it 3 times.

client/chatroomwidget.cpp
```diff
@@ -658,6 +661,70 @@ void ChatRoomWidget::quote(const QString& htmlText)
...
+        m_messageModel->roleNames().key("eventType")).toString();
+    const QString eventId = m_messageModel->data(m_messageModel->index(index, 0),
```
KitsuneRal:
```suggestion
    const auto eventId = m_messageModel->index(index, 0)->data(
```

client/chatroomwidget.cpp
```diff
@@ -658,6 +661,70 @@ void ChatRoomWidget::quote(const QString& htmlText)
     m_chatEdit->insertPlainText(sendString);
 }
 
+void ChatRoomWidget::showMenu(const QString& hoveredLink, bool showingDetails,
+                              int index)
+{
+    const QString eventType = m_messageModel->data(m_messageModel->index(index, 0),
+        m_messageModel->roleNames().key("eventType")).toString();
```
KitsuneRal: Gosh. That's a certain omission in the current code. Can you please move `EventRoles` from `messageeventmodel.cpp` to the header file (inside the class), and use things like `MessageEventModel::EventTypeRole`?

client/chatroomwidget.cpp
```diff
@@ -658,6 +661,70 @@ void ChatRoomWidget::quote(const QString& htmlText)
...
+    const QString eventId = m_messageModel->data(m_messageModel->index(index, 0),
+        m_messageModel->roleNames().key("eventId")).toString();
+    const auto progressInfo = m_messageModel->data(m_messageModel->index(index, 0),
+        m_messageModel->roleNames().key("progressInfo"))
+            .value<QMatrixClient::FileTransferInfo>();
```
KitsuneRal: This should move below under `if (hasAttachment)` - no need to retrieve it on any event.

client/chatroomwidget.cpp
```diff
@@ -658,6 +661,70 @@ void ChatRoomWidget::quote(const QString& htmlText)
...
+
+    const bool hasAttachment = eventType == "image" || eventType == "file";
+    const bool downloaded = hasAttachment && !progressInfo.isUpload &&
+        progressInfo.completed();
```
KitsuneRal: Same as above, respectively.

client/qml/TimelineItem.qml
```diff
@@ -74,6 +74,37 @@ Item {
...
+        onShowDetails: {
+            if (currentIndex === index) {
+                showingDetails = !showingDetails
+                animation.start()
+            }
+        }
+    }
+
+    SequentialAnimation {
+        id: animation
```
KitsuneRal:
```suggestion
        id: detailsAnimation
```

client/chatroomwidget.cpp
```diff
@@ -662,6 +665,68 @@ void ChatRoomWidget::quote(const QString& htmlText)
...
+    a->setCheckable(true);
+    a->setChecked(showingDetails);
+
+    const auto eventType = modelIndex.data(MessageEventModel::EventTypeRole).toString();
+    const bool hasAttachment = eventType == "image" || eventType == "file";
+    if (hasAttachment)
+    {
+        const auto progressInfo = modelIndex.data(MessageEventModel::SpecialMarksRole)
+            .value<QMatrixClient::FileTransferInfo>();
+        const bool downloaded = hasAttachment && !progressInfo.isUpload &&
```
KitsuneRal: `hasAttachment` is known to be `true` - you don't need it in the expression here.

client/qml/TimelineItem.qml
```diff
@@ -74,6 +74,37 @@ Item {
             shownChanged(true);
     }
 
+    property bool showingDetails
+    signal showDetails
```
KitsuneRal: Sorry, forgot one more thing - you don't need this signal anymore.

PR #505, opened 2019-02-23
Add login warning dialog for logging into the same account

### LINKED ISSUE #369: When logged in twice under the same account, room entries cannot be distinguished
An accidental manifestation of this can be found in https://matrix.org/blog/2018/06/22/this-week-in-matrix-2018-06-22/.

krkk:
Closes #369

### DIFF
--- a/client/dialog.cpp
+++ b/client/dialog.cpp
@@ -95,6 +95,7 @@ void Dialog::buttonClicked(QAbstractButton* button)
     switch (buttons->buttonRole(button))
     {
         case QDialogButtonBox::AcceptRole:
+        case QDialogButtonBox::YesRole:
             if (validate())
             {
                 if (statusLabel)
@@ -107,6 +108,7 @@ void Dialog::buttonClicked(QAbstractButton* button)
             load();
             break;
         case QDialogButtonBox::RejectRole:
+        case QDialogButtonBox::NoRole:
             reject();
             break;
         default:
--- a/client/mainwindow.cpp
+++ b/client/mainwindow.cpp
@@ -760,7 +760,27 @@ void MainWindow::showLoginWindow(const QString& statusMessage)
         account.sync();
 
         showFirstSyncIndicator();
-        addConnection(connection, dialog.deviceName());
+
+        QString deviceName = dialog.deviceName();
+        const auto it = std::find_if(connections.cbegin(), connections.cend(),
+            [connection] (Connection* c) {
+                return c->userId() == connection->userId();
+            });
+
+        if (it != connections.cend())
+        {
+            int ret = QMessageBox::warning(this,
+                tr("Logging in into a logged in account"),
+                tr("You're trying to log in into an account that's "
+                   "already logged in. Do you want to continue?"),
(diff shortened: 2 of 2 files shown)


client/mainwindow.cpp
```diff
@@ -760,7 +761,26 @@ void MainWindow::showLoginWindow(const QString& statusMessage)
         account.sync();
 
         showFirstSyncIndicator();
-        addConnection(connection, dialog.deviceName());
+
+        for (auto a: connections)
```
KitsuneRal: Ideally, instead of an explicit `for()` it would be great to do a `std::find()`:
```suggestion
        const auto it = std::find(connections.cbegin(), connections.cend(),
                [a] (Connection* c) { return a->userId() == c->userId(); });
        if (it != connections.cend())
```
If you hate STL algorithms, at least enclose `connections` into `qAsConst` to prevent COW-detaching (see, e.g., https://stackoverflow.com/questions/35811053/using-c11-range-based-for-loop-correctly-in-qt).

client/mainwindow.cpp
```diff
@@ -760,7 +761,26 @@ void MainWindow::showLoginWindow(const QString& statusMessage)
...
+        {
+            if (a->userId() == connection->userId())
```
KitsuneRal: With `std::find()` this becomes unneeded, of course.

client/mainwindow.cpp
```diff
@@ -760,7 +761,26 @@ void MainWindow::showLoginWindow(const QString& statusMessage)
...
+                Dialog warningDialog(tr("Logging in into a logged in account"),
+                                     QDialogButtonBox::Yes|QDialogButtonBox::No,
+                                     this, Dialog::NoStatusLine);
+                auto* label = new QLabel(
+                    tr("You're trying to log in into an account that's "
+                       "already logged in. Do you want to continue?"));
+                warningDialog.addWidget(label);
+
+                if (warningDialog.exec())
+                    deviceName += "-" + connection->deviceId();
```
KitsuneRal: It's really a nitpick but I believe you can `break;` after that (unless rewritten via `std::find()`, see above).

client/mainwindow.cpp
```diff
@@ -746,6 +746,7 @@ void MainWindow::showLoginWindow(const QString& statusMessage)
     {
         auto connection = dialog.releaseConnection();
         AccountSettings account(connection->userId());
+        QString deviceName = dialog.deviceName();
```
KitsuneRal: Can you move it down, closer to where it's actually used? It's not C so we can declare variables anywhere in the block (and I prefer it to grouping variable definitions).

client/dialog.cpp
```diff
@@ -95,6 +95,7 @@ void Dialog::buttonClicked(QAbstractButton* button)
     switch (buttons->buttonRole(button))
     {
         case QDialogButtonBox::AcceptRole:
+        case QDialogButtonBox::YesRole:
```
KitsuneRal: Despite my comment below, this is useful anyway; so please keep it regardless and I'm fine if this sneaks in with the PR.

PR #520, opened 2019-03-03
MainWindow: fix Open room dialog

krkk:


KitsuneRal:
Thanks but I'm afraid this is overly verbose and besides leaks `QComboBox` in cases when there's one connection. I pushed the fix independently (sorry for scrapping this).

Issue #513, opened 2019-03-02
Use colour-coding for user disambiguation


KitsuneRal:
This is much less intrusive than the current (recommended by The Spec, heh) way of attaching MXID to the name.

KitsuneRal:
Closed by #528.

PR #528, opened 2019-03-04
Use colour-coding for user disambiguation

a-andreyev:
Contributes to issue #513 
I like this HSL-based appoarch so tried to help (implemented similar for r0kk3rz Sailfish OS version), but feel to reject/correct :)

Checked with different KDE color schemes:

Imgur

### DIFF
--- a/client/models/userlistmodel.cpp
+++ b/client/models/userlistmodel.cpp
@@ -22,6 +22,8 @@
 #include <QElapsedTimer>
 #include <QtCore/QDebug>
 #include <QtGui/QPixmap>
+#include <QtGui/QPalette>
+#include <QtWidgets/QApplication>
 
 #include <connection.h>
 #include <room.h>
@@ -106,6 +108,16 @@ QVariant UserListModel::data(const QModelIndex& index, int role) const
         if (!user->bridged().isEmpty())
             tooltip += "<br>" + tr("Bridged from: %1").arg(user->bridged());
         return tooltip;
+    }
+
+    if (role == Qt::ForegroundRole)
+    {
+        // FIXME: boilerplate with TimelineItem.qml:57
+        return QColor::fromHslF(user->hueF(),
+                                1 - QApplication::palette().color(QPalette::Window).saturationF(),
+                                -0.7 * QApplication::palette().color(QPalette::Window).lightnessF() + 0.9,
+                                QApplication::palette().color(QPalette::ButtonText).alphaF()
+                                );
     }
 
     return QVariant();
--- a/client/qml/TimelineItem.qml
+++ b/client/qml/TimelineItem.qml
@@ -54,27 +54,12 @@ Item {
         (["state", "notice", "other"].indexOf(eventType) >= 0) ?
                 disabledPalette.text : defaultPalette.text
     readonly property string authorName: room && room.roomMembername(author.id)
-    function stringToHue(str) {
-        var hash = 0
-        if ((str).length === 0)
-            return hash
-        for (var i = 0; i < (str).length; i++) {
-            hash = (str).charCodeAt(i) + ((hash << 5) - hash)
-            hash = hash & hash
-        }
-        return Math.abs(hash)/2147483647
-    }
-    function stringToColor(str) {
-        if (str) {
-            return Qt.hsla(stringToHue(str),
-                           (1-defaultPalette.window.hslSaturation),
-                           /* contrast but not too heavy: */
-                           (-0.6*defaultPalette.window.hslLightness + 0.9),
-                           textColor.a)
-        }
-        return textColor
-    }
-    readonly property color fancyColor: stringToColor(authorName)
+    // FIXME: boilerplate with models/userlistmodel.cpp:115
+    readonly property string authorColor: Qt.hsla(userHue,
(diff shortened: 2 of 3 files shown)


KitsuneRal:
Thanks for the start! And yes, HSL-based calculation is certainly the way to go. We have a few more tasks ahead though :)
- [ ] Calculate colour in the library (https://github.com/QMatrixClient/libqmatrixclient/issues/296)
- [ ] Colour users in the user list as well
- [ ] Provide a switch to use colour disambiguation instead of full-name disambiguation

a-andreyev:
current results screenshot:

Imgur

KitsuneRal:
Now that we have the lib PR, it's time to rebase this PR on the newest `master` :)

client/models/userlistmodel.cpp
```diff
@@ -108,6 +110,15 @@ QVariant UserListModel::data(const QModelIndex& index, int role) const
         return tooltip;
     }
 
+    if (role == Qt::TextColorRole)
```
KitsuneRal: I'd rather we use a custom role; `TextColorRole` is about the "main" text colour and is too easy to confuse with the message colour.

client/qml/TimelineItem.qml
```diff
@@ -46,14 +46,35 @@ Item {
...
+        if (str) {
+            return Qt.hsla(author.hueF,
+                           (1-defaultPalette.window.hslSaturation),
+                           /* contrast but not too heavy: */
+                           (-0.7*defaultPalette.window.hslLightness + 0.9),
+                           textColor.a)
+        }
+        return textColor
+    }
+    readonly property color fancyColor: stringToColor(authorName)
```
KitsuneRal: You don't need all this remake here now that the library gives you everything necessary - just get that custom role I mentioned above.

client/qml/TimelineItem.qml
```diff
@@ -156,7 +177,7 @@ Item {
                     actionEvent ? Text.AlignRight : Text.AlignLeft
                 elide: Text.ElideRight
 
-                color: textColor
+                color: fancyColor
```
KitsuneRal: ...and insert that role name here.

client/qml/TimelineItem.qml
```diff
@@ -240,7 +261,7 @@ Item {
                     // FIXME: The text is clumsy and slows down creation
                     text: (actionEvent && !xchatStyle ?
                            ("<a href='#mention' style='text-decoration:none;color:\"" +
-                                    defaultPalette.text + "\"'><b>" +
+                                    fancyColor + "\"'><b>" +
```
KitsuneRal: ...and here.

client/models/userlistmodel.cpp
```diff
@@ -22,6 +22,8 @@
 #include <QElapsedTimer>
 #include <QtCore/QDebug>
 #include <QtGui/QPixmap>
+#include <QPalette>
+#include <QApplication>
```
KitsuneRal: Qt module names, please.

a-andreyev:
rebased and testing, getting black color for now (checking what versrion I've compiled)

client/models/userlistmodel.cpp
a-andreyev: not sure how to connect `TextColorRole` with custom role for a  `QTableView` at `UserListDock` (widget-based ui part), not changed for now

a-andreyev:
working now, but I have a boilerplate code for final color calculation from hue: for `UserListDock` widget c++ part and `TimelineItem` qml part. Not sure yet where's better place to provide shared solution :thinking:

client/models/userlistmodel.cpp
KitsuneRal: Ah, we don't have any custom roles there yet. Ok, keep it as is now.

KitsuneRal:
At the moment, all CI pipelines are breaking because you seem to have forgotten to add the newest lib to your commit. In order to do that, you have to make sure the contents under `lib/` subdirectory are at the right commit (basically - the newest commit in `master`) and then commit the subdirectory itself.

client/models/userlistmodel.cpp
KitsuneRal: Ok, there's still one thing to fix here: according to Qt docs, `TextColorRole` is obsolete, and `ForegroundRole` should be used instead.

KitsuneRal:
> working now, but I have a boilerplate code for final color calculation from hue: for `UserListDock` widget c++ part and `TimelineItem` qml part. Not sure yet where's better place to provide shared solution 🤔

Well, a fancy way of doing it would be to introduce a Quaternion-specific user class (`QuaternionUser`) instead of the generic `QMatrixClient::User` (`Connection` allows you to customise the class it uses to create user objects) and provide `QuaternionUser::color()` that also backs a `Q_PROPERTY` so that it could be referred to from QML. If that sounds a bit too much for you, just leave the boilerplate - we'll get back to it sometime later.

a-andreyev:
> If that sounds a bit too much for you, just leave the boilerplate - we'll get back to it sometime later.

Probably a bit too much but sounds interesting :slightly_smiling_face: Agree to leave it for now and save the task as an issue for example

Issue #535, opened 2019-03-13
Consider configurable style of quoting


KitsuneRal:

KitsuneRal:
Closed by #543.

PR #543, opened 2019-03-15
Message quoting

### LINKED ISSUE #509: Message quoting (plaintext)
Replies are not quite an easy feat, [as it turned out](https://github.com/QMatrixClient/libqmatrixclient/issues/245#issuecomment-466869313); quotes are much easier (just take the plain body and insert `> ` after each newline).

krkk:
Closes #509, #535

TODO: better description
https://github.com/QMatrixClient/Quaternion/blob/d3f745ceff64cdac48b895269c425fe1f8e8c614/client/mainwindow.cpp#L277

### DIFF
--- a/client/chatroomwidget.cpp
+++ b/client/chatroomwidget.cpp
@@ -35,6 +35,7 @@
 #endif
 #include <QtCore/QRegularExpression>
 #include <QtCore/QStringBuilder>
+#include <QtCore/QLocale>
 
 #include <events/roommessageevent.h>
 #include <csapi/message_pagination.h>
@@ -615,6 +616,42 @@ void ChatRoomWidget::onMessageShownChanged(const QString& eventId, bool shown)
             if (indicesOnScreen.erase(pos) == indicesOnScreen.end())
                 reStartShownTimer();
     }
+}
+
+void ChatRoomWidget::quote(const QString& htmlText)
+{
+    QMatrixClient::SettingsGroup sg { QStringLiteral("UI") };
+    const auto type = sg.get<int>("quote_type");
+    const auto defaultStyle = "> \\1\n";
+    const auto defaultRegex = "(.+)(?:\n|$)";
+    auto style = sg.get<QString>("quote_style");
+    auto regex = sg.get<QString>("quote_regex");
+
+    if (style.isEmpty())
+        style = defaultStyle;
+    if (regex.isEmpty())
+        regex = defaultRegex;
+
+    QTextDocument document;
+    document.setHtml(htmlText);
+    QString sendString;
+
+    switch (type)
+    {
+        case 0:
+            sendString = document.toPlainText()
+                .replace(QRegularExpression(defaultRegex), defaultStyle);
+            break;
+        case 1:
+            sendString = document.toPlainText()
+                .replace(QRegularExpression(regex), style);
+            break;
+        case 2:
+            sendString = QLocale().quoteString(document.toPlainText()) + "\n";
+            break;
+    }
+
+    m_chatEdit->insertPlainText(sendString);
 }
 
 void ChatRoomWidget::reStartShownTimer()
--- a/client/mainwindow.cpp
+++ b/client/mainwindow.cpp
@@ -261,6 +261,27 @@ void MainWindow::createMenu()
         }
     });
 
+    viewMenu->addAction(QIcon::fromTheme("format-text-blockquote"),
+        tr("Edit quote style"), [this]
+    {
+        QMatrixClient::SettingsGroup sg { "UI" };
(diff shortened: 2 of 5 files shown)


client/mainwindow.cpp
```diff
@@ -261,6 +261,44 @@ void MainWindow::createMenu()
         }
     });
 
+    viewMenu->addAction(QIcon::fromTheme("format-text-blockquote"),
+        tr("Edit quote style"), [this]
```
KitsuneRal: I'd probably not want to bring this to UI, not for now at least. I don't think many people will make use of it, and those who do want to use it may look up the documentation (that is, README.md for now) and edit the format in the configuration directly.

What I would probably be interested in is a switch between `> ` and the locale-specific quoting style (see `QLocale::quoteString()`; also, for reference, here's [a mind-numbing pile of numbers](https://code.woboq.org/qt5/qtbase/src/corelib/tools/qlocale_data_p.h.html#locale_data) it uses to select a quote delimiter for a given locale). On UNIX systems, Qt selects it basing on `LC_MESSAGES`.

I don't consider customising delimiters beyond these two options to be a valid use case for any more or less common (even tech-savvy) user but I'm fine to support it on the level of under-the-hood configuration (maybe as the third option, not exposed in the UI - and for that reason we might just provide a means to specify a regex/replacement pair in the config file, such as `(.*)\n` / <code>&#8225; \1\n</code> or `(.*)` / <code>&#8222;\1&#8221;</code> ).

client/chatroomwidget.cpp
```diff
@@ -617,6 +617,33 @@ void ChatRoomWidget::onMessageShownChanged(const QString& eventId, bool shown)
     }
 }
 
+void ChatRoomWidget::quote(const QString& htmlText)
+{
+    static const auto SettingsKey = QStringLiteral("quote_style");
+    static const auto SettingsType = QStringLiteral("quote_type");
```
KitsuneRal: `QStringLiteral` is a macro that makes a string in static storage; you don't need to separately save it to a static variable. In other words, `SettingsKey` and `SettingsType` can be inlined.

client/mainwindow.cpp
```diff
@@ -407,6 +445,13 @@ void MainWindow::createMenu()
 
 void MainWindow::loadSettings()
 {
+    QMatrixClient::SettingsGroup uiSg("UI");
+    if (!uiSg.contains("quote_style"))
+    {
+        uiSg.setValue("quote_style", "> %1");
```
KitsuneRal: Better provide a default value to `get<>` invocations in `ChatRoomWidget::quote()` - it's more robust against things like deleting pieces of configuration while Quaternion is running (there's no guarantee that `QSettings` are read once at startup - actually, quite the opposite).

client/chatroomwidget.cpp
```diff
@@ -617,6 +618,34 @@ void ChatRoomWidget::onMessageShownChanged(const QString& eventId, bool shown)
...
+    document.setHtml(htmlText);
+    QString sendString;
+
+    if (type == 0)
+    {
+        sendString = document.toPlainText().replace(QRegularExpression(regex), style);
+    } else
+    {
+        QLocale loc;
+        sendString = loc.quoteString(document.toPlainText()) + "\n";
```
KitsuneRal: `loc` can be inlined without losing readability. `QLocale()` is a well-documented idiom for the default locale.

client/mainwindow.cpp
```diff
@@ -261,6 +261,24 @@ void MainWindow::createMenu()
...
     });
 
+    viewMenu->addAction(QIcon::fromTheme("format-text-blockquote"),
+        tr("Edit quote style"), [this]
+    {
+        QMatrixClient::SettingsGroup sg { "UI" };
+        const auto type = sg.get<int>("quote_type");
+
+        QStringList list;
+        list << tr("Default (regex)") << tr("Locale's default");
```
KitsuneRal: I totally dig the brevity you achieved; but I'd still probably distinguish between "Markdown (prepend each line with >)" and "Custom (apply regex from the config file)". "Default (regex)" without having a regex around causes quite a confusion.
I'd also probably display what "Locale's default" means by adding something like `+ ' ' + QLocale().quoteString("Example quote")`.

client/chatroomwidget.cpp
```diff
@@ -25,6 +25,7 @@
 #include <QtWidgets/QAction>
 #include <QtWidgets/QFileDialog>
 #include <QtWidgets/QApplication>
+#include <QLocale>
```
KitsuneRal:
```suggestion
#include <QtCore/QLocale>
```
And even better, move this down to group with other QtCore #includes.

PR #547, opened 2019-03-16
MainWindow: completions in the 'Open Room' dialog

krkk:


client/mainwindow.h
```diff
@@ -139,12 +146,17 @@ class MainWindow: public QMainWindow
...
          * \param initialConn initially selected account, if there are several
+         * \param completionType - type of completion
          * \param prompt - dialog box title
          * \param label - the text next to the identifier field (e.g., "User ID")
          * \param actionName - the text on the accepting button
          */
-        Locator obtainIdentifier(Connection* initialConn, const QString& prompt,
-                                 const QString& label, const QString& actionName);
+        Locator obtainIdentifier(Connection* initialConn,
+                                 CompletionType completionType,
```
KitsuneRal: This is very cool but can we use `QFlags<CompletionType>` as a parameter type here? See the next comment about 'Open room'.

client/mainwindow.cpp
```diff
@@ -189,7 +190,7 @@ void MainWindow::createMenu()
     openRoomAction = viewMenu->addAction(QIcon::fromTheme("document-open"),
         tr("Open room..."), [this] {
             resolveLocator(obtainIdentifier(
-                    currentRoom ? currentRoom->connection() : nullptr,
+                    currentRoom ? currentRoom->connection() : nullptr, Room,
```
KitsuneRal: I'd be happy to see `Room|User` here because "Open room" can actually open direct chats with specific users as well.

PR #548, opened 2019-03-16
RoomListDock: remember state of groups

### LINKED ISSUE #499: Remember expanded/collapsed state of groups in the room list
The code is there but doesn't work for some reason, so it's a bug rather than a feature.

krkk:
Closes #499

### DIFF
--- a/client/roomlistdock.cpp
+++ b/client/roomlistdock.cpp
@@ -155,7 +155,10 @@ RoomListDock::RoomListDock(MainWindow* parent)
         }
     });
     connect( model, &RoomListModel::groupAdded, this, [this] (int pos) {
-        view->expand(model->index(pos, 0));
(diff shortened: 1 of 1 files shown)


client/roomlistdock.cpp
```diff
@@ -155,7 +155,10 @@ RoomListDock::RoomListDock(MainWindow* parent)
         }
     });
     connect( model, &RoomListModel::groupAdded, this, [this] (int pos) {
-        view->expand(model->index(pos, 0));
+        QMatrixClient::SettingsGroup sg { QStringLiteral("UI/RoomsDock") };
+        QString group = model->roomGroupAt(model->index(pos, 0)).toString();
```
```suggestion
        const auto group = model->roomGroupAt(model->index(pos, 0)).toString();
```

PR #579, opened 2019-04-18
Audio playback

krkk:
preview


KitsuneRal:
I will review it (possibly tomorrow) but meanwhile I'm just curious - why not use `MediaPlayer` component and cover all bases in one go?

KitsuneRal:
The original branch was deleted.

PR #580, opened 2019-04-20
Hyperlink users

### LINKED ISSUE #191: Make member nicknames hyperlinked before sending
This will in particular improve UX between Q and Riot, as Riot will make those nice pills out of matrix.to hyperlinks

krkk:
Closes #191


client/chatroomwidget.cpp
```diff
@@ -346,7 +346,33 @@ QString ChatRoomWidget::doSendInput()
 
     if (!text.startsWith('/'))
     {
-        m_currentRoom->postPlainText(text);
+        QRegExp MxIdRegExp(QStringLiteral(
```
KitsuneRal: 1. `QRegExp` is an anachronism from Qt4; please use `QRegularExpression` instead.
2. We already have a (less precise) regexp for user MXIDs earlier in this method (`UserIdRE`) - maybe replace that one with yours?

client/chatroomwidget.cpp
```diff
@@ -346,7 +346,33 @@ QString ChatRoomWidget::doSendInput()
...
+        }
+
+        for (int pos = 0; (pos = MxIdRegExp.indexIn(htmlText, pos)) != -1; pos += replaced.count())
+        {
+            replaced =
+                QStringLiteral(R"(%1<a href="https://matrix.to/#/%2">%3</a>)")
+                .arg(MxIdRegExp.cap(1), MxIdRegExp.cap(2),
+                     m_currentRoom->roomMembername(MxIdRegExp.cap(2)).toHtmlEscaped());
+            htmlText.replace(pos, MxIdRegExp.matchedLength(), replaced);
+        }
```
KitsuneRal: I admittedly didn't take time to fully understand what you're doing here but it seems to me like a regex replacement. In which case - why not use the stock `QString::replace` that accepts a regular expression, along the same lines as done in `lib/util.cpp` for linkification?

client/chatroomwidget.cpp
krkk: > 2. We already have a (less precise) regexp for user MXIDs earlier in this method (`UserIdRE`) - maybe replace that one with yours?

I don't think we can replace that.

`UserIdRE` has position characters where mine doesn't, so replacing it might cause false positives
(e.g. `/invite @bob:example.com!` would match but won't send an invite because `argString` it's not a valid user ID as a whole).

We could replace `argString` with `UserIdRE.match(argString).captured()` there, but I think saying that the user ID is incorrect is better than splitting the user IDs and taking the first one.

KitsuneRal (approved):

PR #589, opened 2019-05-13
ChatRoomWidget: remove one extra newline in quotes

krkk:


KitsuneRal:
That's unfortunately not extra. This is to provide compatibility with Markdown that needs this extra line to mark the end of citation. There should be a configuration option when Quaternion supports Markdown but in the meantime I'd prefer to keep the "extra" newline as the default, as long as discontent users can change that in the configuration.

krkk:
You're right, I forgot, but when we quote a message that has more than one line, it will also add an empty line between the quoted lines (which I don't think is the intended effect).

KitsuneRal:
Ouch, no it's not.

PR #602, opened 2019-06-17
Timeline: scroll faster with mouse wheel

rpallai:
> `var yDelta = wheel.angleDelta.y / 120 * 100`

I don't like this fixed multiplier but even so it's a good start and works well for me.

### DIFF
--- a/client/qml/Timeline.qml
+++ b/client/qml/Timeline.qml
@@ -262,6 +262,21 @@ Rectangle {
                 contentY = Math.min(originY + contentHeight - height,
                                     contentY + height)
             }
+            function onWheel(wheel) {
+                if (wheel.angleDelta.x == 0) {
+                    var yDelta = wheel.angleDelta.y / 120 * 100
+
+                    if (yDelta > 0) {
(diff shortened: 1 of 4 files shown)


rpallai:
The new patch doesn't mess up horizontal wheel scroll of oversized textFields.

rpallai:
It doesn't mess cursor shape over the timeline.

rpallai:
The new patch follows the previous approach but provides a more complete implementation.

rpallai:
It seemed like a low-hanging fruit but it isn't.

The previous approach required all content elements to tile the message with mouse areas which isn't sustainable. Instead of this tiling I decided to use of "fallback" areas at the highest level. It's easier, more complete and transparent for content widgets.

The newly introduced type called `TimelineMouseArea` will be used more extensivly in my behaviour changing patches in the future. This is why I don't avoid this new type here.

rpallai:
PR #606 contains this along with other depending mouse handling improvements. I pushed as a whole to save time on review cycles.

PR #605, opened 2019-06-23
Turn settings menu to a dialog window

### LINKED ISSUE #553: Trade most of the Settings menu for a Preferences dialog window

rpallai:
I hope Qt Designer is not banned.

Closes #553.

### DIFF
--- a/client/settingsdialog.cpp
+++ b/client/settingsdialog.cpp
@@ -1,0 +1,120 @@
+/**************************************************************************
+ *                                                                        *
+ * Copyright (C) 2019 Roland Pallai <dap78@magex.hu>                      *
+ *                                                                        *
+ * This program is free software; you can redistribute it and/or          *
+ * modify it under the terms of the GNU General Public License            *
+ * as published by the Free Software Foundation; either version 3         *
+ * of the License, or (at your option) any later version.                 *
+ *                                                                        *
+ * This program is distributed in the hope that it will be useful,        *
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of         *
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the          *
+ * GNU General Public License for more details.                           *
+ *                                                                        *
+ * You should have received a copy of the GNU General Public License      *
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.  *
+ *                                                                        *
+ **************************************************************************/
+
+#include "settingsdialog.h"
+#include "mainwindow.h"
+#include <settings.h>
+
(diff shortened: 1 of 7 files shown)


KitsuneRal (changes requested):
Qt Designer is certainly not banned, and I agree that using it for a preferences window is reasonable. What I can't accept though is that on my machine the preferences dialog is too small for the contents, with scrollbars in both directions even though there's enough room on the screen to make the dialog larger. Can you please somehow bind the size of the dialog to its contents up to a certain limit?

One more suboptimal thing: I understand that tabs are almost inevitable in the long term. But having both tabs and groups looks really over the top. I'd probably favour tabs over bigger captions. Or, if you prefer everything to be on a single page (I fully agree it's better from usability perspective) then hide the tab strip until it's really needed.

rpallai:
The window size issue is fixed if I'm not mistaken, please test. The tab strip has been hidden.

KitsuneRal:
Still not :(
Capture

rpallai:
It must be good now.

KitsuneRal:
Finally got to this - thanks, it works pretty much as expected now. Would you please rebase it to the latest master (and, if you don't mind, add a tab with fonts configuration)?

rpallai:
Rebased and font settings have been exposed. I think adding tabs is unnecessary for now as the settings window is still pretty short.

client/settingsdialog.cpp
```diff
@@ -0,0 +1,195 @@
...
+    connect(timeline_layout, QOverload<int>::of(&QComboBox::currentIndexChanged), this, [this](int index) {
+        auto new_layout = timeline_layout->itemData(index).toString();
+        SettingsGroup("UI").setValue("timeline_style", new_layout);
+        emit timelineChanged();
+    });
+
+    // timelineFont
+    const auto curTimelineFontFamily = SettingsGroup("UI").value("Fonts/timeline_family", "").toString();
+    timelineFontFamily->insertItem(0, tr("system default"));
+    if (curTimelineFontFamily == "") {
```
KitsuneRal:
```suggestion
    if (curTimelineFontFamily.isEmpty()) {
```

KitsuneRal (commented):
Thanks a lot! I have one major piece of feedback regarding the font part - instead of a pair of combo boxes it would probably be better to use a font dialog (despite the dialog-over-dialog ugliness). The reason here is that some fonts have a few weights that can be equally fitting (light, regular and semibold, e.g.), and 6-18 range for font sizes might not be quite right for some extreme cases. In other words, a platform-native font dialog would certainly cover the use case better than the font configuration approximation with two combo boxes. What do you think?

Issue #300, opened 2018-01-31
deselect when selecting code of another message


krombel:
As you can see you do not know which piece is currently selected. 
![deselction does not work](https://matrix.org/_matrix/media/v1/download/msgs.tk/UORoTuuWTSynUbaQzkpWpSKS)

When you select another part of the same message deselection works but not for different messages

KitsuneRal:
Selection as well as copy-pasting from the timeline is a complete mess at the moment and it's unlikely to change any time soon :( Basically, there's a choice between enabling that weird selection or disabling selection altogether. I preferred to enable at least this. To fix selection properly one should either change the container used to render the timeline from QML ListView to something more accommodating or do very non-trivial hackarounds, effectively reimplementing selection and copy-pasting. I don't know which is harder, and unlikely to have time even to research it in the nearest couple of months.

PR #606, opened 2019-06-27
Timeline: mouse handling improvements

rpallai:


rpallai:
> if I wasn't careful enough and selected too much or too little, I have to reselect it

How can you correct your selection currently? Using the shift + arrow keys on the keyboard?

rpallai:
I reworked text selection, now the selected text:
1. remains selected after selection ended (as before)
2. immediately copied into global mouse selection buffer (as before)
3. could be copied to the clipboard with ctrl+c (as before)
4. gets unselected when new selection starts (new feature)

The only feature I have to reimplement is correction of the selection by shift + click. To tell you truth I wasn't aware of this capability until you've mentioned it.

rpallai:
> The only feature I have to reimplement is correction of the selection by shift + click.

It's also done. Please test.

client/kchatedit.cpp
```diff
@@ -222,6 +222,12 @@ void KChatEdit::keyPressEvent(QKeyEvent *event)
             d->forwardHistory();
         }
         break;
+    case Qt::Key_C:
```
KitsuneRal: You're stepping into a pretty dangerous territory of catching up with 40 years of confusion and fighting over key shortcuts for common operations. Some platforms still assume Ctrl+Insert instead of Ctrl+C to copy text - let alone all crazy folks with customised shortcuts. In short: I really recommend using `matches(QKeySequence::Copy)` here instead.

client/chatroomwidget.cpp
```diff
@@ -725,6 +730,14 @@ void ChatRoomWidget::showMenu(int index, const QString& hoveredLink,
     menu.exec(QCursor::pos());
 }
 
+void ChatRoomWidget::setGlobalSelectionBuffer(QString text)
+{
+    if (QApplication::clipboard()->supportsSelection())
+        QApplication::clipboard()->setText(text, QClipboard::Selection);
```
KitsuneRal: Won't it happen even without this code?

client/qml/TimelineTextEditSelector.qml
```diff
@@ -0,0 +1,50 @@
...
+            x += textEdit.flickableItem.contentX
+            y += textEdit.flickableItem.contentY
+        }
+        var hasSelection = textEdit.selectionEnd > textEdit.selectionStart
+        if (hasSelection && controller.getModifierKeys() & Qt.ShiftModifier) {
+            textEdit.moveCursorSelection(textEdit.positionAt(x, y), selectionMode)
+        } else {
+            textEdit.cursorPosition = textEdit.positionAt(x, y)
+            if (chatView.textEditWithSelection)
+                chatView.textEditWithSelection.deselect()
```
KitsuneRal: Since you invested so much into reimplementing selection and copying - maybe you would support selection across messages, or at least fix #300 along the way?

client/chatroomwidget.cpp
rpallai: No, it won't. I tested it.

client/qml/TimelineTextEditSelector.qml
rpallai: Actually, it closes #300 as is. I just wasn't aware of the fact that #300 is exists. It has been mentioned.

client/kchatedit.cpp
```diff
@@ -204,6 +204,13 @@ QSize KChatEdit::sizeHint() const
 
 void KChatEdit::keyPressEvent(QKeyEvent *event)
 {
+    if (event->type() == QEvent::KeyPress || event->type()== QEvent::ShortcutOverride) {
+        if (static_cast<QKeyEvent *>(event)->matches(QKeySequence::Copy)) {
```
KitsuneRal: Isn't it `QKeyEvent` already?

rpallai: True. I should've paid more attention. Fixed.

KitsuneRal:
Turned out I was a bit hasty. Your code prevents hyperlinks from working.

KitsuneRal:
Also, after spending some more time I now see that the original scrolling is actually about 3 times more precise (it's also slower but not as much) at least on my setup. A case for adjustable sensitivity is quite prominent.

rpallai:
> Also, after spending some more time I now see that the original scrolling is actually about 3 times more precise (it's also slower but not as much) at least on my setup. A case for adjustable sensitivity is quite prominent.

What do you mean under more precise scrolling exactly? Wasn't your jump length constant per every each wheelEvent? Perhaps `wheel.pixelDelta` is supported by your system?
On my Linux, the original scrolling speed was about 16px / wheelEvent constantly which was way too slow. The new 100px / wheelEvent is a pretty big jump.

Anyway, adjustable sensitivity is indeed important, I'll push that into the preferences window as it will be available (#605).

KitsuneRal:
That's the contention point: for me, 16px/notch is almost good enough (maybe a bit slowish). So it's really just about different tastes.

rpallai:
Oh, I overcomplicated this issue, I thought "sensitivity" is non-linear to "speed" - thanks for clarification. I noticed fa8ded0 restored your speed preference. It's okay, I'll make it configurable as #605 settles down. The hard part is done by now. :confetti_ball:

KitsuneRal:
Ouch, I didn't really mean to restore it so unceremoniously :( And of course let's make it configurable after merging #605.

With respect to sensitivity - we can make it non-linear but it's yet another Pandora's box. From the UX perspective - very few people would even bother so much to precisely tune it. What I would introduce instead is adding the shuttle-dial style of treating wheel events - with notches controlling the speed rather than distance. That would effectively enable auto-scrolling (imagine speed-reading your timeline without having to constantly turn the wheel to show the next piece).

PR #608, opened 2019-06-27
Roomlistdock: restore room switching by Enter

rpallai:
It could be pretty useful for keyboard users. It was cancelled by cd88a867 for unknown reasons.


client/roomlistdock.cpp
```diff
@@ -105,7 +105,7 @@ RoomListDock::RoomListDock(MainWindow* parent)
 
     static const auto Expanded = QStringLiteral("expand");
     static const auto Collapsed = QStringLiteral("collapse");
-//    connect( view, &QTreeView::activated, this, &RoomListDock::rowSelected );
+    connect( view, &QTreeView::activated, this, &RoomListDock::rowSelected );
```
```suggestion
    connect( view, &QTreeView::activated, this, &RoomListDock::rowSelected ); // See #608
```

PR #612, opened 2019-07-03
Timeline: restore `Item#message` height shrink ability

rpallai:
This got broken by dbda393 which introduced circular dependency on height: `Item#message` height was dependent on `childrenRect.height` but it included a `MouseArea` which was dependent on `parent.height`.

It turned out by 87b18d9 via changing `discardButton` type from `ActiveLabel` to `ToolButton` which is taller and forces sent messages to shrink when it gets invisible (in normal circumstances, shortly after it appears).

### DIFF
--- a/client/qml/TimelineItem.qml
+++ b/client/qml/TimelineItem.qml
@@ -331,27 +331,20 @@ Item {
 
                 TimelineMouseArea {
                     anchors.fill: parent
-                    acceptedButtons: Qt.MiddleButton
-
-                    onClicked: {
-                        if (textFieldImpl.hoveredLink)
-                            controller.resourceRequested(
-                                textFieldImpl.hoveredLink)
-                    }
-                }
-
-                TimelineMouseArea {
-                    anchors.fill: parent
-                    acceptedButtons: Qt.RightButton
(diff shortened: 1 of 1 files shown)


KitsuneRal:
Now that I look once again at the code, I start wondering if we really need that many `TimelineMouseArea`s.
1. Does it make sense to consolidate similar mouse areas into one under the top-level `Item` with `anchors.fill: fullMessage` instead of having one mouse area per each piece of the message? (and then this PR would be unnecessary in its current form)
2. We now have 3 `TimelineMouseArea`s under `textField` - I strongly suspect we can consolidate them into one.

rpallai:
> 1. Does it make sense to consolidate similar mouse areas into one under the top-level `Item` with `anchors.fill: fullMessage` instead of having one mouse area per each piece of the message? (and then this PR would be unnecessary in its current form)

You are absolutely right. The reason of having different MouseAreas was historical but the code was reworked several times since and it isn't necessary anymore. I forgot to consolidate them. Thanks!

> 2. We now have 3 TimelineMouseAreas under textField - I strongly suspect we can consolidate them into one.

I agree, it's a practical simplification. Pushed.

PR #614, opened 2019-07-07
Support custom chatting font

rpallai:
Using the same font for the timestamp, author, message items and
input box creates a unified look.

The timestamp's font is not smaller anymore. The main reason behind
this decision is that it's challenging to perfectly align the
timestamp with different sized font to the first line of the
message which could be bothering. I don't think it's worth it.

Co-Authored-By: Denis Danilov <ddanilov@users.noreply.github.com>

### DIFF
--- a/client/kchatedit.cpp
+++ b/client/kchatedit.cpp
@@ -16,10 +16,26 @@
  */
 
 #include "kchatedit.h"
+#include <settings.h>
 
 #include <QDebug>
 #include <QGuiApplication>
 #include <QKeyEvent>
+
+static QTextDocument* makeDocument()
+{
+    auto textDocument = new QTextDocument();
+
+    const auto fontFamily = Quotient::Settings().value("UI/Timeline/font_family");
+    const auto fontPointSize = Quotient::Settings().value("UI/Timeline/font_pointSize");
+    if (fontFamily.isValid() && fontPointSize.isValid() && fontPointSize.toReal() > 0) {
+        auto font = QFont(fontFamily.toString());
+        font.setPointSizeF(fontPointSize.toReal());
+        textDocument->setDefaultFont(font);
+    }
+
+    return textDocument;
+}
 
 class KChatEdit::KChatEditPrivate
 {
@@ -34,7 +50,7 @@ public:
     // History always ends with a placeholder string that is initially empty
     // but may be filled with tentative input when the user entered something
     // and then went out for history.
-    QVector<QTextDocument*> history { 1, new QTextDocument() };
+    QVector<QTextDocument*> history { 1, makeDocument() };
     int index = 0;
     int maxHistorySize = 100;
 };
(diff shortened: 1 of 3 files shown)


client/kchatedit.h
```diff
@@ -124,4 +124,12 @@ class KChatEdit : public QTextEdit
     Q_DISABLE_COPY(KChatEdit)
 };
 
+class KChatEditDocument : public QTextDocument
```
KitsuneRal: That's a pretty heavy solution if all you need is to customise `QTextDocument`. Can't you rewrite it as a factory function and replace all `new QTextDocument()` with calls to it?

client/kchatedit.cpp
```diff
@@ -16,11 +16,23 @@
...
 
 #include "kchatedit.h"
+#include <settings.h>
 
 #include <QDebug>
 #include <QGuiApplication>
 #include <QKeyEvent>
 
+
+KChatEditDocument::KChatEditDocument(QWidget *parent)
```
KitsuneRal: This could be turned to the body of that factory with minimal changes. By the way, you don't even need to make it a member of `KChatEdit`, since you don't really use its state/interface. Something like `static QTextDocument makeDocument()` inside the .cpp file should be enough.

client/kchatedit.cpp
```diff
@@ -16,11 +16,24 @@
...
 #include <QKeyEvent>
 
+static QTextDocument* makeDocument()
+{
+    auto textDocument = new QTextDocument();
+
+    const auto fontFamily = Quotient::Settings().value("UI/Timeline/font_family");
+    const auto fontPointSize = Quotient::Settings().value("UI/Timeline/font_pointSize");
+    if (fontFamily.isValid() && fontPointSize.isValid())
+        textDocument->setDefaultFont(QFont(fontFamily.toString(), fontPointSize.toReal()));
```
KitsuneRal: `QFont::QFont()` only accepts `int`, not `double`; your value is truncated to `int` here. You should create a font and then use `setPointSizeF()`. Also, `isValid()` for the font size might be slightly too weak - maybe use `canConvert<qreal>()`?

KitsuneRal:
Oh, and please extend the custom font towards the room header.

client/kchatedit.cpp
rpallai: What do you think of using `fontPointSize.toReal() > 0` instead of `canConvert<qreal>`?

KitsuneRal: Well, it's a slight bit heavier but more reliable, I agree. Let's do that.

KitsuneRal:
As a postscriptum (and as an ultimate solution for the issue above) - it might be more consistent in the end to just bluntly call `QApplication::setFont()` in `main.cpp` and override the font where really necessary instead.

KitsuneRal:
I got carried away and implemented the two-stage system of font settings: application-wide and timeline-wide. A side-effect of this is that I consolidate all font settings under `UI/Fonts`: in particular, the timeline font settings are (going to be) in `UI/Fonts/timeline_*` instead of `UI/Timeline/font_*` - sorry for disrupting your settings.

PR #621, opened 2019-08-12
TimelineItem: fix unpressable scrollbar in the details pane

### LINKED ISSUE #619: Scrollbar in the event source is unpressable
@rpallai, you might want to look into this. The mouse area apparently overrides presses/clicks over the scrollbar. Perhaps it shouldn't cover the details pane at all?

rpallai:
We don't care about lost input focus when interacting with the
details pane anymore because it is a rare event. This careless
attitude is the best fix for the unpressable scrollbar there.

Got broken by 4fbcbb3e, closes #619


KitsuneRal:
So be it.

PR #625, opened 2019-08-14
README.md: add blank line before lists

rpallai:
It is not required by the specification but most editors are
implemented this way. Both syntax look the same on GitHub
rendering.

In short, I want to use my Ghostwriter! :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,7 @@ messages; number and date formats will be following the setting though. Version
 0.0.9.4 gains German, Polish, and Russian translations.
 
 Quaternion stores its configuration in a way standard for Qt applications. It will read and write the configuration in the user-specific location (creating it if non-existent) and will only read the system-wide location with reasonable defaults if the configuration is nowhere to be found.
+
 - Linux:
   - system-wide: `$XDG_CONFIG_DIR/Quotient/quaternion` or
     `/etc/xdg/Quotient/quaternion`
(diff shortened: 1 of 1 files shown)


KitsuneRal:
Are most editors implemented this way? You should probably file a bug for Ghostwriter ;) All or almost all editors I used handled lists without the preceding blank line totally fine.

If that's the only serious quirk, I'm ok to merge it. But I really don't want to go on a slippery path of trying to make the text look right in all the Markdown renderers out there.

KitsuneRal:
Actually, now that I contemplate over this, I think that moving the docs over to Asciidoc is not such a bad idea...

rpallai:
> Are most editors implemented this way? You should probably file a bug for Ghostwriter ;) All or almost all editors I used handled lists without the preceding blank line totally fine.

I tried the following editors under Linux and none of them rendered it well: ReText, Ghostwriter, CuteMarkEd. To tell you the truth, I haven't found an editor which supports lists without preceding blank line. It's weird because these editors support GitHub flavour of markdown in theory and actually there aren't any rendering quirks beyond this.

What editor are you using?

KitsuneRal:
Well, I'm on Windows and used a Markdown plugin for Notepad++ as well as Markdown Navigator plugin for CLion.

If it's that bad on Linux, I'll merge this one. Still wonder if converting to .adoc is a more reliable way to stay compatible...

rpallai:
> [...], I'll merge this one.

💕🙂

> Still wonder if converting to .adoc is a more reliable way to stay compatible...

I haven't used Asciidoc yet, but now I'm happy with this markdown documentation too.

PR #626, opened 2019-08-15
[WIP] Remodel desktop notifications

rpallai:
This is a request for comments to improve desktop notifications.

First of all, I'm not a big fan of the `client` mode but showing the content of received messages in the popup is a must and I wasn't able to find a better solution. The old default behaviour is available with the `popup_mode=server` and `popup_tweaks=highlight+,intrusive` settings.

Todo:
- [ ] discuss the approach
- [ ] test on Windows and OSX (help needed)
- [ ] remove useless debug code

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -161,6 +161,7 @@ set(quaternion_SRCS
     client/chatedit.cpp
     client/chatroomwidget.cpp
     client/systemtrayicon.cpp
+    client/popupnotifier.cpp
     client/models/messageeventmodel.cpp
     client/models/userlistmodel.cpp
     client/models/roomlistmodel.cpp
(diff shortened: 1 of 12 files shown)


KitsuneRal:
For better or worse, the client-side triggering of notifications is the way to go, simply because it's the only working way with E2EE. I also agree that showing the content preview is a wanted feature. In order to avoid code duplication in events rendering, though, I'd like to move the event visualisation code to the library (this will also help to remove code duplication between Spectral and Quaternion). I wanted to do it for quite a bit of time already but never got even close to starting. If you feel brave enough to get into this rabbit hole, let's discuss how to do it in the library in an extensible way.

rpallai:
> For better or worse, the client-side triggering of notifications is the way to go, simply because it's the only working way with E2EE.

Good point, I didn't take E2EE into consideration. Client-side triggering is here to stay.

> In order to avoid code duplication in events rendering, though, I'd like to move the event visualisation code to the library

I take your point but I'm mainly motivated in practical usability improvements rather than code refactoring so I'm afraid this is still waiting for you. Would you merge this PR with this small volume of code duplication?

Does it work on Windows?

PR #628, opened 2019-08-16
Fix ChatRoomWidget.readMarkerOnScreen tracking with local echo

rpallai:
The symptom of this bug is the activity detector stopped working
after a message was sent to a room with enabled local echo (the
default).


KitsuneRal:
Oh. Of course. Thank you!

PR #629, opened 2019-08-20
Timeline: Pg{Up,Down}: do not jump more than a page

rpallai:
Additionally, little bit of overlap should be kept between pages
for eye comfort.


rpallai:
> I buy the idea but I'm not a fan of absolute values. [...] can we make it something like, say, 98% of the height but no less than 5 pixels?

You're right, but I think it always should be close to the height of a typical line of text. That feels natural. See my latest push.

PR #630, opened 2019-08-20
Timeline: get rid of some broken math with Qt 5.10

rpallai:
Scrolling to the last page using page down wasn't pixel-perfect in
many cases. The old algorithm couldn't cope with hidden events, on
demand rendered content, etc. Fortunately, we don't have to do this
math since Qt 5.10.

I understand if you postpone this PR because Qt 5.10 is lacking in some major distributions.


KitsuneRal:
Yeah, it will be great - when we move over to Qt 5.12 (in a year, maybe?).

Issue #616, opened 2019-07-15
No lib SHA in About


KitsuneRal:
CMake shows lib SHA but it doesn't seem to go through to UI.

PR #635, opened 2019-09-06
MainWindow: show lib SHA in About

krkk:
Fixes: #616


KitsuneRal:
@KitsuneRal facepalms. I looked several times at this code, never seeing the obvious. Thank you!

PR #642, opened 2019-11-17
maybe-read-timer setting

ddanilov:
replace the hardcoded value by a parameter setting

### DIFF
--- a/client/chatroomwidget.cpp
+++ b/client/chatroomwidget.cpp
@@ -756,7 +756,7 @@ void ChatRoomWidget::reStartShownTimer()
             indexToMaybeRead >= indicesOnScreen.back())
         return;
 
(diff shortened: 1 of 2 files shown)


client/chatroomwidget.cpp
```diff
@@ -756,7 +756,7 @@ void ChatRoomWidget::reStartShownTimer()
             indexToMaybeRead >= indicesOnScreen.back())
         return;
 
-    maybeReadTimer.start(1000, this);
+    maybeReadTimer.start(QSettings().value("UI/maybe_read_timer", 1000).value<int>(), this);
```
KitsuneRal: Can you make a `Settings` (instead of `QSettings`) object static and use `Settings::get<>`? Otherwise things look good.

PR #645, opened 2019-11-23
opacity read marker

ddanilov:
option to make already read messages visually different from unread ones
(by adding transparency)

### DIFF
--- a/client/models/messageeventmodel.cpp
+++ b/client/models/messageeventmodel.cpp
@@ -57,6 +57,7 @@ QHash<int, QByteArray> MessageEventModel::roleNames() const
     roles[EventResolvedTypeRole] = "eventResolvedType";
     roles[RefRole] = "refId";
     roles[ReactionsRole] = "reactions";
+    roles[DimmerReadMarkerRole] = "dimmerReadMarker";
     return roles;
 }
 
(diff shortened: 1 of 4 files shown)


client/models/messageeventmodel.cpp
```diff
@@ -156,12 +157,17 @@ void MessageEventModel::changeRoom(QuaternionRoom* room)
...
+            this, [this] (QString fromEventId, QString toEventId) {
             refreshEventRoles(
                 std::exchange(lastReadEventId,
                               m_currentRoom->readMarkerEventId()),
                 {ReadMarkerRole});
             refreshEventRoles(lastReadEventId, {ReadMarkerRole});
+            // the iterators are revert iterators
+            const auto fromEventIdIt = m_currentRoom->findInTimeline(fromEventId);
+            const auto toEventIdIt = m_currentRoom->findInTimeline(toEventId);
+            auto refreshDimming = [this](auto& timeLineEvent){ refreshEventRoles(timeLineEvent.event()->id(), {DimmerReadMarkerRole}); };
```
KitsuneRal, line 169: Did you think of trying to calculate the index from the read marker instead of introducing a new role? Like passing the last read index and then calculating the background depending on whether this index is less or more than the last read?

ddanilov: > Did you think of trying to calculate the index from the read marker instead of introducing a new role? 

no, because a role is the only way I know right now to pass something from c++ to qml layer

> Like passing the last read index and then calculating the background depending on whether this index is less or more than the last read?

actually it is not the background but the opacity of the text that is dimmed.
and then its calculation should happen inside qml when passing only the  the last read index?
and how to pass it there w/o a role?

ddanilov:
I'm going to learn more about Qml :) , indeed it looks like that could be done somewhere around onShown() function or so...

Issue #581, opened 2019-04-25
Feature Request: allow hiding of other events


peterhoeg:
### Description

It is currently possible to [hide join/leave events](https://github.com/QMatrixClient/Quaternion/issues/248), but it would be great to be able to hide others too:

- [ ] bans
- [ ] name changes
- [ ] avatar changes
- [ ] setting of room aliases
- [ ] and possibly others - basically anything *not* a regular message

### Steps to reproduce

n/a

### Version information

n/a

Issue #411, opened 2018-09-17
Optionally hide unknown events


KitsuneRal:

KitsuneRal:
Basically a matter of adding one more option to the View menu and processing it in the message event model.

PR #654, opened 2020-03-13
MessageEventModel: allow hiding some types of events

krkk:
Invites, ban events, display name changes, avatar changes, and unknown events can be now hidden.

Closes: #411
Closes: #487
Closes: #581

### DIFF
--- a/client/mainwindow.cpp
+++ b/client/mainwindow.cpp
@@ -260,6 +260,12 @@ void MainWindow::createMenu()
         tr("Show/hide meaningless activity (join-leave pairs and redacted events between)"),
         QStringLiteral("show_spammy")
     );
+    addTimelineOptionCheckbox(
+        showEventsMenu,
+        tr("Un&known event types"),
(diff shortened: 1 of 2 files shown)


client/mainwindow.cpp
```diff
@@ -201,20 +201,56 @@ void MainWindow::createMenu()
...
     addTimelineOptionCheckbox(
         showEventsMenu,
         tr("&Redacted events"),
         tr("Show redacted events in the timeline as 'Redacted'"
            " instead of hiding them entirely"),
         QStringLiteral("show_redacted")
     );
+    addTimelineOptionCheckbox(
+        showEventsMenu,
+        tr("Display na&me changes"),
```
KitsuneRal, line 235:
```suggestion
        tr("Changes in display na&me"),
```
to avoid the confusion

KitsuneRal:
Actually, while we're at it - can you also cover #411 in this PR?

Issue #596, opened 2019-05-21
Aggregations, message editing, reactions


ptman:
Are these features planned? What's needed? Are there already issues for them?

KitsuneRal:
They are being worked on; experimental support in the library is already there, and some initial things (either reactions or message editing - whichever comes easier) will be in Quaternion soon (c)

grahamperrin:
For me, Quaternion shows that there were reactions, but not _what_ those reactions were: 

https://user-images.githubusercontent.com/192271/103240169-75a22b80-4947-11eb-8a68-f016196b4df6.png

```text
mowa219-gjp4-8570p% date ; uname -v                     
Mon 28 Dec 2020 19:59:56 GMT
FreeBSD 13.0-CURRENT #74 r368589: Sun Dec 13 07:55:46 GMT 2020     root@mowa219-gjp4-8570p:/usr/obj/usr/src/amd64.amd64/sys/GENERIC-NODEBUG 
mowa219-gjp4-8570p% pkg query '%o %v %R' quaternion
net-im/quaternion 0.0.9.5.b1 FreeBSD
mowa219-gjp4-8570p% 
```

KitsuneRal:
@grahamperrin this is because the font doesn't have the glyph for the reaction. Unfortunately, Quaternion doesn't allow to pick a different font for reactions (such as Noto Emoji) and default fonts on most platforms (Arial/Helvetica and such) don't have a lot of emoji glyphs.

ecsgh:
> For me, Quaternion shows that there were reactions, but not _what_ those reactions were:
> 
> https://user-images.githubusercontent.com/192271/103240169-75a22b80-4947-11eb-8a68-f016196b4df6.png

Same here with newest quaternion and newest Quotient.
When will this be fixed?

KitsuneRal:
I'm exactly one developer working on Quaternion as an unpaid, spare time project. So the answer to "when" will invariably be - when you submit a pull request, or when I get to it.

Now if we get to technical things - the problem is mainly due to:
1. Apparent unavailability of emoji characters in your font, as mentioned above.
2. Inability of Qt to get out of stalemate and pick a different font for emojis: https://bugreports.qt.io/browse/QTBUG-85744

A font with at least fallback (black-and-white) emoji characters will help with (1). I don't know what can help with (2) - engaging with The Qt Project doesn't seem to help. I'm not very fond of working it around and forcing the font on emoji characters on Quaternion side, but PRs are welcome.

PR #655, opened 2020-03-14
Show reactions

krkk:
...and add ability to react to a message with already existing reactions.

screenshot

Contributes to #596

### DIFF
--- a/client/chatroomwidget.cpp
+++ b/client/chatroomwidget.cpp
@@ -745,7 +745,8 @@ void ChatRoomWidget::reactionButtonClicked(const QString& eventId, const QString
         m_currentRoom->relatedEvents(eventId, EventRelation::Annotation());
 
     for (const auto& a : annotations) {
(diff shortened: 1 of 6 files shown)


client/chatroomwidget.cpp
```diff
@@ -737,6 +738,23 @@ void ChatRoomWidget::showMenu(int index, const QString& hoveredLink,
...
 }
 
+void ChatRoomWidget::reactionButtonClicked(const QString& eventId, const QString& key)
+{
+    using namespace Quotient;
+    const auto& annotations =
+        m_currentRoom->relatedEvents(eventId, EventRelation::Annotation());
+
+    for (const auto& a : annotations) {
+        if (eventCast<const ReactionEvent>(a)->relation().key == key
```
KitsuneRal, line 748: You'll crash dereferencing a null pointer if anything except `ReactionEvent` hits the list of annotations. That's not possible as yet but may become any time soon.

PR #658, opened 2020-03-15
user setting to disable keychain

ddanilov:


client/mainwindow.cpp
```diff
@@ -585,6 +591,12 @@ bool MainWindow::saveAccessTokenToFile(const AccountSettings& account,
 bool MainWindow::saveAccessTokenToKeyChain(const AccountSettings& account,
                                            const QByteArray& accessToken, bool writeToFile)
 {
+    if (!Quotient::SettingsGroup("UI").value("use_keychain", true).toBool())
```
KitsuneRal:
```suggestion
    if (!Quotient::Settings().value("UI/use_keychain", true).toBool())
```
A tiny bit more optimal. `SettingsGroup` is better when you intend to use different values from the same group, `Settings` is just fine for one-off reads/writes.

client/mainwindow.cpp
```diff
@@ -475,6 +475,12 @@ QByteArray MainWindow::loadAccessTokenFromFile(const AccountSettings& account)
 #ifdef USE_KEYCHAIN
 QByteArray MainWindow::loadAccessTokenFromKeyChain(const AccountSettings& account)
 {
+    if (!Quotient::SettingsGroup("UI").value("use_keychain", true).toBool())
```
KitsuneRal:
```suggestion
    if (!Quotient::Settings().value("UI/use_keychain", true).toBool())
```

client/mainwindow.cpp
```diff
@@ -1016,27 +1028,30 @@ void MainWindow::logout(Connection* c)
...
-                       << qPrintable(job.errorString());
-            if (job.error() != QKeychain::Error::NoBackendAvailable &&
-                job.error() != QKeychain::Error::NotImplemented &&
-                job.error() != QKeychain::Error::OtherError)
-            {
-                QMessageBox::warning(this, tr("Couldn't delete access token"),
-                                     tr("Quaternion couldn't delete the access "
-                                        "token from the keychain."),
-                                     QMessageBox::Close);
+    if (Quotient::SettingsGroup("UI").value("use_keychain", true).toBool())
```
KitsuneRal:
```suggestion
    if (Quotient::Settings().value("UI/use_keychain", true).toBool())
```

Issue #659, opened 2020-03-15
ChatRoomWidget::roomSettingsRequested() is referenced but does not exist


krkk:
https://github.com/quotient-im/Quaternion/blob/a7d6c90f3fdf3b683d1b6cef54b9d1a86c7bbdd7/client/qml/Timeline.qml#L185-L189

PR #660, opened 2020-03-16
ChatRoomWidget: define roomSettingsRequested()

krkk:
Closes: #659

### DIFF
--- a/client/mainwindow.cpp
+++ b/client/mainwindow.cpp
@@ -90,6 +90,8 @@ MainWindow::MainWindow()
              this, &MainWindow::openResource);
     connect( chatRoomWidget, &ChatRoomWidget::joinRequested,
              this, &MainWindow::joinRoom);
+    connect( chatRoomWidget, &ChatRoomWidget::roomSettingsRequested,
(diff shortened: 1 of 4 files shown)


client/mainwindow.cpp
```diff
@@ -90,6 +90,14 @@ MainWindow::MainWindow()
              this, &MainWindow::openResource);
     connect( chatRoomWidget, &ChatRoomWidget::joinRequested,
              this, &MainWindow::joinRoom);
+    connect( chatRoomWidget, &ChatRoomWidget::roomSettingsRequested,
+             this, [this]
+            {
+                auto* dlg = new RoomSettingsDialog(currentRoom, this);
```
KitsuneRal: What about moving out [the hashmap of room settings dialogs](https://github.com/quotient-im/Quaternion/pull/660/files#diff-506d4ee3de7ac421bb6551a5a365f756L294) from that lambda to `MainWindow` class (no more as static) and using [`summon()`](https://github.com/quotient-im/Quaternion/pull/660/files#diff-506d4ee3de7ac421bb6551a5a365f756L134) on its `[currentRoom]` value?

PR #667, opened 2020-04-01
Fix resizing of the text selection

krkk:
Creating a new selection in the same message resulted in:
```
qrc:/qml/TimelineTextEditSelector.qml:22: TypeError: Property 'getModifierKeys' of object ChatRoomWidget(0x5616507a9ca0) is not a function
```

### DIFF
--- a/client/chatroomwidget.h
+++ b/client/chatroomwidget.h
@@ -52,7 +52,7 @@ class ChatRoomWidget: public QWidget
         bool pendingMarkRead() const;
 
         QStringList findCompletionMatches(const QString& pattern) const;
-        Qt::KeyboardModifiers getModifierKeys() const;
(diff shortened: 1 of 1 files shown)


client/qml/TimelineTextEditSelector.qml
```diff
@@ -19,7 +19,7 @@ MouseArea {
             y += textEdit.flickableItem.contentY
         }
         var hasSelection = textEdit.selectionEnd > textEdit.selectionStart
-        if (hasSelection && controller.getModifierKeys() & Qt.ShiftModifier) {
+        if (hasSelection && controller.getModifierKeys & Qt.ShiftModifier) {
```
KitsuneRal, line 22: This has to be a function still.

PR #678, opened 2020-06-07
Add image attaching from the clipboard

krkk:


client/chatedit.cpp
```diff
@@ -34,6 +36,10 @@ void ChatEdit::keyPressEvent(QKeyEvent* event)
         triggerCompletion();
         return;
     }
+    if (event->matches(QKeySequence::Paste) && !QApplication::clipboard()->image().isNull()) {
```
KitsuneRal, line 39: This misses on pasting via the right mouse button. The Qt docs mention the correct way of handling this (see `QTextEdit::paste()`):

> To change the behavior of this function, i.e. to modify what `QTextEdit` can paste and how it is being pasted, reimplement the virtual `canInsertFromMimeData()` and `insertFromMimeData()` functions.

What I'd like to see is `ChatEdit` figuring out that the MIME data contain an image, and emitting a signal indicating that attaching a file has been requested - with the MIME data as a parameter. The chat room widget then could process this uniformly for both drag-n-dropping an image and pasting it - and further on extend to other types of data, e.g. audio.

client/chatroomwidget.h
```diff
@@ -97,6 +98,7 @@ class ChatRoomWidget: public QWidget
         MessageEventModel* m_messageModel;
         QuaternionRoom* m_currentRoom;
         ImageProvider* m_imageProvider;
+        QTemporaryFile* m_temporaryClipboardImage;
```
KitsuneRal, line 101:
```suggestion
        QTemporaryFile* m_fileToAttach;
```

client/chatroomwidget.cpp
```diff
@@ -140,6 +147,21 @@ ChatRoomWidget::ChatRoomWidget(QWidget* parent)
...
                 : selectedText);
     });
+    connect(m_chatEdit, &ChatEdit::pasteClipboardImage, this, [=] {
+        if (m_temporaryClipboardImage->isOpen() || m_currentRoom == nullptr)
+            return;
+
+        m_temporaryClipboardImage->open();
+
+        QApplication::clipboard()->image()
+            .save(m_temporaryClipboardImage, "PNG");
```
KitsuneRal, line 157: Following on the comment above, you'll have MIME data here so you won't need to access the clipboard.

client/chatroomwidget.cpp
```diff
@@ -140,6 +148,20 @@ ChatRoomWidget::ChatRoomWidget(QWidget* parent)
                 ? m_chatEdit->textCursor().selectedText()
                 : selectedText);
     });
+    connect(m_chatEdit, &ChatEdit::insertFromMimeDataRequested,
+            this, [=] (const QMimeData* source) {
+        if (m_fileToAttach->isOpen() || m_currentRoom == nullptr)
+            return;
+        m_fileToAttach->open();
+
+        qvariant_cast<QImage>(source->imageData()).save(m_fileToAttach, "PNG");
```
KitsuneRal, line 157: I wonder how this will treat SVG files ;) anyway, SVG can be covered separately.

PR #699, opened 2020-09-09
MainWindow: fix logout action

krkk:


KitsuneRal (approved):
Argh, that's what happens when I do shortcuts at testing. Thanks!

PR #721, opened 2020-12-28
Start porting to QQC2

ognarb:
image

There is a few regressions and a few bug fixes. This is still WIP and I need to figure a clean way to use only stuff from Qt 5.9 or Qt 5.12.


KitsuneRal:
I got carried away on the weekend and accidentally ported Timeline.qml to (Qt 5.9 version of) QQC2 :) Sorry but it seems you'll have to rebase this PR on top of the current master.

PR #724, opened 2021-01-06
Fix tr call missing

ognarb:
found it while copying a bit of code :)


KitsuneRal (approved):

KitsuneRal:
Huh - thanks for contributing back :)

PR #729, opened 2021-01-21
scroll to reader marker

### LINKED ISSUE #257: Jump to last read message

ddanilov:
resolves #257

### DIFF
--- a/client/qml/ScrollToButton.qml
+++ b/client/qml/ScrollToButton.qml
@@ -1,0 +1,22 @@
+import QtQuick.Controls 2.2
+
+RoundButton {
+    height: settings.defaultText.height * 2
+    width: height
+    hoverEnabled: true
+    opacity: visible * (0.7 + hovered * 0.2)
+
+    display: Button.IconOnly
+    icon.color: defaultPalette.buttonText
+
+    AnimationBehavior on opacity {
+        NormalNumberAnimation {
+            easing.type: Easing.OutQuad
+        }
+    }
+    AnimationBehavior on anchors.bottomMargin {
+        NormalNumberAnimation {
+            easing.type: Easing.OutQuad
+        }
+    }
+}
--- a/client/qml/Timeline.qml
+++ b/client/qml/Timeline.qml
@@ -623,34 +623,47 @@ Rectangle {
         }
     }
 
-    RoundButton {
+    ScrollToButton {
         id: scrollToBottomButton
+
         anchors.right: parent.right
         anchors.bottom: parent.bottom
         anchors.rightMargin: width * 1.5
-        anchors.bottomMargin: chatView.atYEnd ? -height : height/2
-        height: settings.defaultText.height * 2
-        width: height
-        hoverEnabled: true
-        opacity: (!chatView.atYEnd) * (0.7 + hovered * 0.2)
+        anchors.bottomMargin: visible ? 0.5 * height : -height
 
-        display: Button.IconOnly
+        visible: !chatView.atYEnd
+
         icon {
             name: "go-bottom"
             source: "qrc:///scrolldown.svg"
-            color: defaultPalette.buttonText
         }
 
         onClicked: {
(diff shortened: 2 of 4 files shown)


client/qml/Timeline.qml
```diff
@@ -653,4 +653,35 @@ Rectangle {
...
+        anchors.bottomMargin: v ? 0.5*height : -3*height
+        height: settings.defaultText.height * 2
+        width: height
+        hoverEnabled: true
+        opacity: v*(0.7 + hovered * 0.2)
+        property bool v: messageModel.readMarkerVisualIndex > chatView.indexAt(chatView.contentX, chatView.contentY)
+
+        display: Button.IconOnly
+        icon {
+            name: "go-top"
```
KitsuneRal, line 671: What would be a fallback if the icon is not there?

client/qml/Timeline.qml
```diff
@@ -653,4 +653,35 @@ Rectangle {
...
+        id: scrollToReaderMarkerButton
+        anchors.right: parent.right
+        anchors.bottom: scrollToBottomButton.top
+        anchors.rightMargin: width * 1.5
+        anchors.bottomMargin: v ? 0.5*height : -3*height
+        height: settings.defaultText.height * 2
+        width: height
+        hoverEnabled: true
+        opacity: v*(0.7 + hovered * 0.2)
+        property bool v: messageModel.readMarkerVisualIndex > chatView.indexAt(chatView.contentX, chatView.contentY)
```
KitsuneRal, line 667: That's a good start but maybe it would also be great to have this button when the read marker is below? Might need to use a different icon for that case though.

KitsuneRal (changes requested):
Thanks for the PR! I was hoping someone would beat me to it :) Now the main problem with it is the amount of copy-pasted code. Can you please move the common part to a separate QML file and use it from `Timeline.qml` twice, configuring the component as necessary?

client/qml/Timeline.qml
ddanilov: I will add one, based on scrolldown.svg

client/qml/Timeline.qml
ddanilov: maybe in a later change, as the first step I would like to keep in one direction only

client/qml/ScrollToButton.qml
```diff
@@ -0,0 +1,36 @@
+import QtQuick.Controls 2.2
+
+RoundButton {
+    anchors.right: parent.right
+    anchors.bottom: parent.bottom
+    anchors.rightMargin: width * 1.5
+    anchors.bottomMargin: visible ? height / 2 : -height
```
KitsuneRal, line 7: Better define anchors directly in the instance - I don't think it can be defined as a "default" for _some_ scroll-to button.

client/qml/ScrollToButton.qml
```diff
@@ -0,0 +1,36 @@
...
+    icon {
+        name: "go-bottom"
+        source: "qrc:///scrolldown.svg"
+        color: defaultPalette.buttonText
+    }
+
+    onClicked: {
+        chatView.positionViewAtBeginning()
+        chatView.saveViewport()
+    }
```
KitsuneRal, line 24: Again, instance-specific behaviour. Ideally, define a custom `signal scrollRequested` or something, and do:
```suggestion
    onClicked: {
        scrollRequested()
        chatView.saveViewport()
    }
```
then implement `onScrollRequested` handler in each instance. Or just entirely define `onClicked` in each instance - that's simpler, even if duplicates the `saveViewport()` call.

KitsuneRal (changes requested):
Thanks, that's better - but I'd prefer the two instances to be equal, not one enjoying default anchoring and signal handling, and the other overriding all that. `ScrollToButton` as an abstract class should be neutral.

KitsuneRal (approved):
Fabulous!

PR #736, opened 2021-02-03
Cache QRegularExpressions in QuaternionRoom::checkForHighlights

nicolasfella:
The function is called a lot and it often rebuilds the same
QRegularExpression since the localUserId doesn't change often.

This is quite wasteful since building a QRegularExpression is quite
expensive (compared to the matching).

Instead save the expressions in a static hash and only build them when
needed.


KitsuneRal (approved):
Makes sense.

PR #742, opened 2021-02-04
Add rich reply support

### LINKED ISSUE #447: Support replying
Depends on https://github.com/QMatrixClient/libqmatrixclient/issues/245.

rpallai:
Closes #447.

### DIFF
--- a/client/chatroomwidget.cpp
+++ b/client/chatroomwidget.cpp
@@ -490,7 +490,7 @@ void ChatRoomWidget::sendMessage(const QString& text, int textType)
         const auto htmlQuote = QStringLiteral(
             "<mx-reply><blockquote><a href=\"%1\">In reply to</a> <a href=\"%2\">%3</a><br />%4</blockquote></mx-reply>"
         ).arg(
-            "https://matrix.to/#/" + m_currentRoom->id() + "/" + referredEventId,
+            Uri(m_currentRoom->id().toUtf8(), referredEventId.toUtf8()).toUrl(Uri::MatrixToUri).toString(),
             Uri(authorUser->id()).toUrl(Uri::MatrixToUri).toString(),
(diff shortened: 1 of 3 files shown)


client/chatroomwidget.cpp
```diff
@@ -446,32 +456,62 @@ void ChatRoomWidget::sendFile()
...
+        const auto htmlQuote = QStringLiteral(
+            "<mx-reply><blockquote><a href=\"%1\">In reply to</a> <a href=\"%2\">%3</a><br />%4</blockquote></mx-reply>"
+        ).arg(
+            Uri(m_currentRoom->id().toUtf8(), referredEventId.toUtf8()).toUrl(Uri::MatrixToUri).toString(),
+            Uri(authorUser->id()).toUrl(Uri::MatrixToUri).toString(),
+            authorUser->displayname(m_currentRoom),
+            citation
+        );
+        // derive plain text fallback
+        const auto plainTextQuote = QLocale().quoteString(QTextDocumentFragment::fromHtml(citation).toPlainText()) + "\n";
```
KitsuneRal, line 499: I know that the spec is especially sadistic in that place but it [tells](https://matrix.org/docs/spec/client_server/latest#stripping-the-fallback) to:

1. Use `> `(basically, Markdown)-style quoting rather than locale-specific quoting
2. Also strip the fallback from the plain-text part.

At least the first part should really be done so that interaction with other clients didn't lead to a pile of replied to payloads accumulated in the body.

KitsuneRal (changes requested):
I really appreciate your heroic effort in implementing that part of the spec (arguably, one of the worst throughout the entire thing) but it's still just a bit not there - specifically in the area of plain text fallback.

rpallai:
Superseded by PR#848

PR #745, opened 2021-02-06
clip the content of room topic inside the scroll view

ddanilov:
otherwise a suffuciently long topic text could expand into the chatView area and
make it unreadable


ddanilov:
for example see rooms
https://view.matrix.org/room/!GNXiKMQpxjFHsivQWJ:chat.weho.st/
https://view.matrix.org/room/!LEdnbfkEStLztPInMH:matrix.org/

KitsuneRal:
Thanks!

PR #747, opened 2021-02-07
RoomListModel: add unread and highlight indicators to tags

krkk:
preview


KitsuneRal (commented):
Thanks - I like the highlighting part definitely but I'm not sure about the number of notifications - would it not be confused with the number of rooms in the group?

KitsuneRal (approved):
Scratch the previous review - we already have the number of rooms there, and a small confusion between the number of rooms with notifications and the number of notifications is probably fine.

PR #770, opened 2021-03-20
ChatEdit: fix ‘Reset formatting’ action status tip not being displayed

krkk:
I had to use this ugly way of declaring an action, because the parent object cannot be specified in `QMenu::addAction()`,  which led to the status tip not being displayed.

Unfortunately, this doesn’t fix the shortcut issue – the QShortcut still has to be in the constructor to trigger the Ctrl+M shortcut.


KitsuneRal (approved):
Heh, indeed. Thanks!

PR #780, opened 2021-07-05
collapsible room topic

ddanilov:
click on room avatar will collapse/uncollapse the room topic in the header

### DIFF
--- a/client/qml/Timeline.qml
+++ b/client/qml/Timeline.qml
@@ -53,6 +53,8 @@ Rectangle {
         radius: 2
         visible: room
 
+        property bool showTopic: true
+
         Image {
             id: roomAvatar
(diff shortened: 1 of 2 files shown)


KitsuneRal (changes requested):
Thanks - that's certainly a very nice idea but can you use a dedicated button or pseudo-hyperlink for that? Clicking on a room avatar is a somewhat concealed and not really expected action for the purpose (I'd rather expect it to open room settings, or the room avatar itself).

ddanilov:
now using more traditional approach, with the context menu

KitsuneRal (approved):
Not sure if a popup menu for this is the best option (it's not exactly discoverable) but that's better than nothing, so let it be like this for now. Thanks!

PR #808, opened 2021-10-23
remove merge chance shield

PiotrZakrzewski:
Merge Chance is going to be deprecated soon


KitsuneRal:
That's unfortunate :( do you know a similar badge to replace it?

PiotrZakrzewski:
> That's unfortunate :( do you know a similar badge to replace it?

I am not aware of an equivalent shield that could replace it, but I also dont keep track of developments in this space that much.

PR #812, opened 2021-11-07
feat: attach file by inserting(pasting/dropping) it into chatedit

shunf4:
Allows attaching a file, if it is inserted (pasted or drag-n-dropped) into message edit box.

I also noted the function `ChatRoomWidget::fileDrop` but it doesn't seem to be used.


KitsuneRal:
`ChatRoomWidget::fileDrop()` invocation has been missing in action since the timeline rework, it seems... Thanks for reinstating the feature!

PR #818, opened 2022-01-29
Systray improvements

Tamaranch:
The first two commits are fixes for what I think are bugs.

The third commit is a minimal change to make the notification system usable in practice (from my point of view), once the two previous fixes are applied.

The fourth commit is an improvement proposal. I have been using the whole thing for a few days now locally.

I'm new to Qt, and my memories of C++ are old (I'm from the C/GTK side :slightly_smiling_face:), so I hope I didn't do too much stupid things or reinvent the wheel.

### DIFF
--- a/client/mainwindow.cpp
+++ b/client/mainwindow.cpp
@@ -591,6 +591,8 @@ void MainWindow::addConnection(Connection* c)
         });
     connect(c, &Connection::loginError, this,
             [this, c](const QString& msg) { reloginNeeded(c, msg); });
+    for (auto* r: c->allRooms())
+        systemTrayIcon->newRoom(r);
     connect(c, &Connection::newRoom, systemTrayIcon, &SystemTrayIcon::newRoom);
     connect(c, &Connection::createdRoom, this, &MainWindow::selectRoom);
     connect(c, &Connection::joinedRoom, this, [this](Room* r, Room* prev) {
--- a/client/systemtrayicon.cpp
+++ b/client/systemtrayicon.cpp
@@ -38,22 +38,22 @@ SystemTrayIcon::SystemTrayIcon(MainWindow* parent)
 
 void SystemTrayIcon::newRoom(Quotient::Room* room)
 {
-    highlightCountChanged(room);
-    connect(room, &Quotient::Room::highlightCountChanged,
-            this, [this,room] { highlightCountChanged(room); });
+    unreadStatsChanged(room);
+    connect(room, &Quotient::Room::unreadStatsChanged,
+            this, [this,room] { unreadStatsChanged(room); });
 }
 
-void SystemTrayIcon::highlightCountChanged(Quotient::Room* room)
+void SystemTrayIcon::unreadStatsChanged(Quotient::Room* room)
 {
     using namespace Quotient;
     const auto mode = Settings().get<QString>("UI/notifications", "intrusive");
     if (mode == "none")
         return;
-    if( room->highlightCount() > 0 ) {
+    if( room->notificationCount() > 0 ) {
         showMessage(
             //: %1 is the room display name
-            tr("Highlight in %1").arg(room->displayName()),
-            tr("%Ln highlight(s)", "", room->highlightCount()));
+            tr("Notification in %1").arg(room->displayName()),
+            tr("%Ln notification(s)", "", room->notificationCount()));
         if (mode != "non-intrusive")
             m_parent->activateWindow();
         connectSingleShot(this, &SystemTrayIcon::messageClicked, m_parent,
--- a/client/systemtrayicon.cpp
+++ b/client/systemtrayicon.cpp
@@ -49,7 +49,7 @@ void SystemTrayIcon::unreadStatsChanged(Quotient::Room* room)
     const auto mode = Settings().get<QString>("UI/notifications", "intrusive");
     if (mode == "none")
         return;
-    if( room->notificationCount() > 0 ) {
+    if (qApp->activeWindow() != nullptr && room->notificationCount() > 0) {
         showMessage(
             //: %1 is the room display name
             tr("Notification in %1").arg(room->displayName()),
--- a/client/systemtrayicon.cpp
+++ b/client/systemtrayicon.cpp
@@ -30,10 +30,14 @@ SystemTrayIcon::SystemTrayIcon(MainWindow* parent)
         showHideAction->setText(visible ? tr("Hide") : tr("Show"));
     });
 
-    setIcon(QIcon::fromTheme(appIconName(), QIcon(":/icon.png")));
+    m_appIcon = QIcon::fromTheme(appIconName(), QIcon(":/icon.png"));
+    m_unreadIcon = QIcon::fromTheme("mail-unread", m_appIcon);
+    m_notified = false;
+    setIcon(m_appIcon);
     setToolTip("Quaternion");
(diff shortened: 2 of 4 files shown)


client/systemtrayicon.cpp
```diff
@@ -89,3 +106,13 @@ void SystemTrayIcon::showHide()
         m_parent->setFocus();
     }
 }
+
+void SystemTrayIcon::focusChanged(QWidget* old, QWidget* now)
+{
+    if (m_notified && old == NULL && qApp->activeWindow() != NULL)
+    {
```
CarlSchwan:
```suggestion
    if (m_notified && old == NULL && qApp->activeWindow() != nullptr) {
```

client/systemtrayicon.cpp
```diff
@@ -89,3 +106,13 @@ void SystemTrayIcon::showHide()
...
+        setIcon(m_appIcon);
+        setToolTip("Quaternion");
```
CarlSchwan, line 105:
```suggestion
        setToolTip(tr("Quaternion"));
```

Tamaranch: There is already one occurrence above (in the constructor) that is not translated: should I add it to the translation too?

CarlSchwan: Best to ask @kitsune :) Generally it's better to have every string in the user interface wrapped in i18n/tr called as some languages transliterate even the 'brand' names

KitsuneRal: We don't translate Quaternion.

client/mainwindow.cpp
```diff
@@ -844,6 +844,8 @@ void MainWindow::addConnection(Connection* c, const QString& deviceName)
         });
     connect(c, &Connection::loginError, this,
             [this, c](const QString& msg) { loginError(c, msg); });
+    for (auto* r: c->allRooms())
+        systemTrayIcon->newRoom(r);
```
KitsuneRal, line 595: At this point `allRooms()` will always return an empty container because the first sync hasn't occurred yet; and during syncs, the tray is updated about new room objects via the `newRoom` signal - the respective `connect()` call is right below here.

client/systemtrayicon.cpp
```diff
@@ -40,29 +40,46 @@ SystemTrayIcon::SystemTrayIcon(MainWindow* parent)
...
 
-void SystemTrayIcon::highlightCountChanged(Quotient::Room* room)
+void SystemTrayIcon::unreadStatsChanged(Quotient::Room* room)
 {
     using namespace Quotient;
     const auto mode = Settings().get<QString>("UI/notifications", "intrusive");
-    if (mode == "none")
+    int nNotifs = 0;
+
+    if (qApp->activeWindow() != nullptr || room->notificationCount() == 0)
```
KitsuneRal, line 56: Highlights and notifications are not exactly the same thing in Matrix. It's a bit complex but basically highlights are supposed to be "louder" than notifications. E.g. a new message in a 1:1 chat normally causes a notification but not a highlight - for a highlight to be triggered, one has to be mentioned. I'm not quite sure if it's best to elevate notifications to system tray popups; but you might have a point.

client/systemtrayicon.cpp
```diff
@@ -40,29 +40,46 @@ SystemTrayIcon::SystemTrayIcon(MainWindow* parent)
...
         return;
-    if( room->highlightCount() > 0 ) {
+
+    for (auto* r: room->connection()->allRooms())
```
KitsuneRal: I'm not sure what you try to do here. The room's connection can be one of several currently active: if I'm logged in with two accounts in Matrix you'll only count notifications from one here. If you want to sum up all currently active notifications (potentially a big lot, unlike highlights - see my other comment) then you should traverse across all connections (stored in `AccountRegistry`).

client/systemtrayicon.cpp
```diff
@@ -89,3 +106,12 @@ void SystemTrayIcon::showHide()
         m_parent->setFocus();
     }
 }
+
+void SystemTrayIcon::focusChanged(QWidget* old, QWidget* now)
```
KitsuneRal:
```suggestion
void SystemTrayIcon::focusChanged(QWidget* old)
```
A nice thing about Qt slots is that you don't need to fully match the signal parameters; unused parameters at the end can be dropped.

client/systemtrayicon.h
```diff
@@ -38,10 +38,14 @@ class SystemTrayIcon: public QSystemTrayIcon
         void newRoom(Quotient::Room* room);
 
     private slots:
-        void highlightCountChanged(Quotient::Room* room);
+        void unreadStatsChanged(Quotient::Room* room);
         void systemTrayIconAction(QSystemTrayIcon::ActivationReason reason);
+        void focusChanged(QWidget* old, QWidget* now);
```
KitsuneRal:
```suggestion
        void focusChanged(QWidget* old);
```

KitsuneRal (changes requested):
There are a few things that could be better here and there but the main question I have for now is whether it's good to trigger on notifications instead of highlights. Given that the loudest ("intrusive") mode focuses the application, a series of notifications from messages in a 1:1 chat may be quite disruptive if they come, say, every minute or even more frequently. Perhaps it's worth handling notifications and highlights differently in that mode specifically. Also, as I wrote below, you seem to mis-accumulate the number of notifications across the application if that's what you tried to do.

client/systemtrayicon.cpp
Tamaranch: It should be good now.

client/mainwindow.cpp
Tamaranch: That's not what I'm seeing: this change does get me into `systemTrayIcon::newRoom()` at startup (so the container doesn't seem empty), and that's necessary so that I'm notified of unread messages from the previous session (I start Quaternion in hidden mode `--hide-mainwindow`).

In fact, if I don't add this initialization, I never see a pass in `systemTrayIcon::newRoom()` for rooms that are already present at startup, so no connection is made to their notification signals.

client/systemtrayicon.cpp
Tamaranch: Oh ok, I see. I seem to have a somewhat peculiar and minimalist use of the software, actually: maintaining a single IRC connection via Matrix on a few fairly low-traffic channels, where I'm almost never mentioned, hence my impression that the notifications weren't working.

So it may be that the current state of the notification system suits the majority of users, and that this PR is not very relevant. I will continue to use this patch for me, as it is suitable for my case, but maybe it is better to just close this PR?

client/mainwindow.cpp
KitsuneRal: Hm, interesting, maybe I misunderstand something in the way things work now...

client/systemtrayicon.cpp
KitsuneRal: You might want to login to Matrix from another client (Element Web, e.g.) once, and tune notifications for those rooms (or even for the whole account) so that all notifications are "loud" (the rightmost setting in respective lines). As for this PR, I think at least some pieces of it are still worthwhile - I really like the icon change, in particular - so perhaps I'll cherry-pick those.

KitsuneRal:
Based on your code I made 584425359082d6bcbe00ec3ee2aa5f0222252f01, and your commits are in the merged trail - thanks a lot for the contribution!

PR #831, opened 2022-09-02
Missing dependency packages for Fedora 36

queria:
Seems two packages were missing when trying to build on Fedora 36:
+ qt5-qtquickcontrols2-devel
+ qt5-linguist

After installing these too I was able to build Quaternion.


KitsuneRal (approved):
Thanks!

PR #836, opened 2022-12-20
Adapt to isReplaced being moved to RoomMessageEvent

TobiasFella:


KitsuneRal (changes requested):
Actually, there's already a branch for `RoomMessageEvent` around line 767 - would you mind sticking one more `if` there instead?

TobiasFella:
Done

KitsuneRal (approved):

PR #838, opened 2022-12-21
Port to libQuotient's keychain and account handling

TobiasFella:
-  At the moment, the "Stay logged in" option is ignored and the connection is always stored in the keychain. I don't think libQuotient currently allows us to have this option
- When logging in, the "You're already logged in with this account; do you want to continue" [not the exact string] dialog was removed for mostly technical reasons; IMO this should be completely blocked before even logging in since there's no point in logging in multiple times with the same connection and libQuotient might explode from doing it :)

Also, I'm not familiar with the codebase, so this is done relatively blindely - please test well.


CMakeLists.txt
```diff
@@ -237,9 +233,7 @@ endif ()
 target_link_libraries(${PROJECT_NAME} Quotient
     ${Qt}::Widgets ${Qt}::Quick ${Qt}::Qml ${Qt}::Gui ${Qt}::Network ${Qt}::QuickControls2 ${Qt}::QuickWidgets)
 
-target_compile_definitions(${PROJECT_NAME} PRIVATE USE_KEYCHAIN) # Legacy, remove in 0.0.96+
-target_link_libraries(${PROJECT_NAME} ${QTKEYCHAIN_LIBRARIES})
-include_directories(${QTKEYCHAIN_INCLUDE_DIR})
+target_link_libraries(${PROJECT_NAME})
```
KitsuneRal: ?

client/accountselector.cpp
```diff
@@ -2,25 +2,29 @@
...
     connect(this, QOverload<int>::of(&QComboBox::currentIndexChanged), this,
             [this] { emit currentAccountChanged(currentAccount()); });
 
-    const auto& accounts = registry->accounts();
+    const auto& accounts = Accounts.accounts();
     for (auto* acc: accounts)
         addItem(acc->userId(), QVariant::fromValue(acc));
 
-    connect(registry, &AccountRegistry::addedAccount, this, [this](Account* acc) {
+    connect(&Accounts, &AccountRegistry::rowsInserted, this, [this](const QModelIndex&, int first, int) {
```
KitsuneRal: What if I insert several accounts at once? There's no such case for now but very well can be.

client/accountselector.cpp
```diff
@@ -2,25 +2,29 @@
...
+        auto acc = Accounts.accounts()[first];
         if (const auto idx = indexOfAccount(acc); idx == -1)
             addItem(acc->userId(), QVariant::fromValue(acc));
         else
             qWarning()
                 << "AccountComboBox: refusing to add the same account twice";
     });
-    connect(registry, &AccountRegistry::aboutToDropAccount, this,
-            [this](Account* acc) {
+    connect(&Accounts, &AccountRegistry::rowsAboutToBeRemoved, this,
```
KitsuneRal, line 27: Same here.

client/mainwindow.h
```diff
@@ -104,7 +102,7 @@ class MainWindow: public QMainWindow, public Quotient::UriResolverBase {
         bool visitNonMatrix(const QUrl& url) override;
 
     private:
-        AccountRegistry accountRegistry;
+        Quotient::AccountRegistry accountRegistry;
```
KitsuneRal: Is it needed, with `Accounts` singleton?

KitsuneRal (changes requested):
Overall looks fairly reasonable, except a few comments and the fact that CI now fails when you build against an external (not a submodule) Quotient.

KitsuneRal (approved):
Thank you!

KitsuneRal:
I forgot to address the "limitations":

> * At the moment, the "Stay logged in" option is ignored and the connection is always stored in the keychain. I don't think libQuotient currently allows us to have this option

At first, I thought that this option should just go because with the current E2EE state (without SSSS and soft logouts) it's all too easy to lose the database. On the other hand, I personally am using it for temporary logins into test accounts... When E2EE is off, I guess it's as easy as making `saveAccessTokenToKeychain()` call optional. With E2EE though, I guess I would opt into an in-memory database instead of storing anything on the disc... Maybe something for 0.8.

>     * When logging in, the "You're already logged in with this account; do you want to continue" [not the exact string] dialog was removed for mostly technical reasons; IMO this should be completely blocked before even logging in since there's no point in logging in multiple times with the same connection and libQuotient might explode from doing it :)

Well, I tested libQuotient to work in this case, before E2EE :) and I see that E2EE doesn't really support it, binding the database to the user account, not to the device. I agree that it's a very exotic case and don't mind dropping it; but we have to add some safeguards in the login procedure then.

PR #851, opened 2023-02-16
TimelineTextEditSelector: enable preventStealing

### LINKED ISSUE #857: Impossible to select text, as display starts scrolling and doesn't clear previous selection
### Description

Someone posts a message into chat and I want to copy the text and paste it into another application. I can start highlighting the text I want to select, but if I have even the slightest vertical movement of the mouse, the selection will be terminated and the screen will start to scroll up/down.  Most of the time I can't even select more than a couple of words without having some slight vertical movement, let alone an entire line, or the entire message that spans multiple lines.

In addition the selection process seems to remember where you previously selected. So if you have a couiple of words selected and want to try again select the text, it starts the selection at where the previous incomplete selection ended, rather than where I've now clicked.

You can also end up having multiple messages selected, as it doesn't seem to clear the old selection on distinct chat messages.

### Steps to reproduce

- Attempt to highlight the text for a single chat message
- Any vertical movement of the mouse terminates the selection at the current pointer location and does vertical scrolling
- After a failed selection, attempt to select the text again, and it won't select from the mouse starting position, instead using end of the previous selection.

### Version information

- **Quaternion version**: quaternion-0.0.95.50~20230106g0315b395-1.fc37.x86_64
- **Qt version**: qt5-qtbase-5.15.8-6.fc37.x86_64
- **Install method**: Distro providqed RPM
- **Platform**: Fedora 37, GNOME shell desktop session

rpallai:
Event stealing cannot be allowed because current text selection gestures conflict with flickable gestures.

### DIFF
--- a/client/qml/TimelineTextEditSelector.qml
+++ b/client/qml/TimelineTextEditSelector.qml
@@ -10,6 +10,7 @@ MouseArea {
 
     anchors.fill: parent
     acceptedButtons: Qt.LeftButton
+    preventStealing: true
 
     onPressed: {
         var x = mouse.x


rpallai:
Actually, flickable gestures continue to work with middle/right mouse button, only left button will be reserved for text selection.

KitsuneRal:
I'm not really a fan of even allowing mouse buttons for dragging/flicking; sadly, we don't really have a good way to discern between taps (on laptops with touchscreens, e.g.) and mouse clicks. Anyway - thanks for this small fix.

PR #852, opened 2023-02-16
TimelineTextEditSelector: fix selection extension

rpallai:
Extension of current selection by shift+click was broken due to
this method wasn't brought to the right class when `TimelineWidget`
was factored out of `ChatRoomWidget`.

Error message:
`qrc:/qml/TimelineTextEditSelector.qml:23: TypeError: Property 'getModifierKeys' of object TimelineWidget(0x24026c0) is not a function`

### DIFF
--- a/client/chatroomwidget.cpp
+++ b/client/chatroomwidget.cpp
@@ -830,8 +830,3 @@ void ChatRoomWidget::textDrop(const QString& text)
 {
     m_chatEdit->insertPlainText(text);
 }
-
-Qt::KeyboardModifiers ChatRoomWidget::getModifierKeys() const
-{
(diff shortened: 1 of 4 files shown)


KitsuneRal (approved):

KitsuneRal (changes requested):
Sorry, I was too hasty with the previous review. Actually, this method is defined in `ChatRoomWidget` but wasn't brought to the right class when `TimelineWidget` was factored out of `ChatRoomWidget`. So would you mind deleting `ChatRoomWidget::getModifierKeys()`, while we're at it?

KitsuneRal (approved):

Issue #849, opened 2023-02-10
Pasting pretty much anything from web pages does not work


rpallai:
It almost always leads to "invalid or unsupported" error message.

In the worst case the plain text version should be inserted.

berrange:
I'm seeing the same problem with pasting pretty much any text from firefox. 

For context, I'm using quaternion-0.0.95.1-3.fc37.x86_64 on Fedora 37 login session with GNOME Desktop plus Wayland.

My workaround is to paste the text from firefox into a terminal or editor, then copy it again and paste into quaternion.

berrange:
FYI, experimenting I see two common scenarios.

The first case which I hit 90% of the time is that the text I'm pasting simply has no HTML elements present, eg because I've just selected some text from the middle of a paragraph

In this case quaternion prints this on the stderr:

```
"Invalid XHTML: Diff selection: mark the radio boxes of the versions to compare and hit enter or the button at the bottom.
Error at char 1: "Start tag expected."
Buffer at error: iff selection: mark the radio boxes of the versions to compare and hit enter or the button at the bottom.
HTML insertion failed at pos 1 with error "Start tag expected."
```

So the HTML parser is not seeing any start tag and thus considers it invalid HTML. 

IMHO this is simply and plainly broken. Any bare text should be considered valid HTML.

The less common case is where the text I've selected in the browser is not well formed because it lacks correct opening/closing tag pairs.  I've see the browser emit such code when selecting text that crosses sections of the page - some opening or closing tags may get omitted depending on where I've started or ended the text selection..

From the POV of parsing the pasted text, it is understandable that the HTML parser be unhappy with this, and it is somewhat tricky to reconcile. None the less, as a user I feel like we should still be able to paste the selected text into a message. Almost any other app successfully pastes the text in this scenario, because they simply ask the original paste selection owner for the text/plain version rather than the text/html version,and the browser is happy to supply the plain text with the mis-matched tags omitted.  I feel like if HTML parsing fails, it would be better to request the selection as plain text and paste that, rather than rejecting it

ghost:
Mismatched tags are also valid in many cases in html, just not in xhtml. The parser basically'd need to parse html and not xhtml :)

KitsuneRal:
@berrange thanks for the investigation. The first case was indeed caused by a blunder and will be fixed shortly. And I agree that reverting to the plain text version with feedback about it in the status bar is a good approach for the second case.

[...]

PR #854, opened 2023-02-17
Chatedit: make default paste mode configurable

rpallai:
Many people like pasting plain text as there is less trouble with
it and does the job. Even Element pastes plain text by default
despite being a rich client. This is why this patch switches to
plain text pasting by default while makes it configurable.

Other version can be pasted by `Ctrl+Shift+V`.

Cc #849

### DIFF
--- a/client/chatedit.cpp
+++ b/client/chatedit.cpp
@@ -77,8 +77,8 @@ void ChatEdit::contextMenuEvent(QContextMenuEvent *event)
     menu->addAction(action);
 
     action = new QAction(QIcon::fromTheme("edit-paste"),
-                         tr(pastePlaintextByDefault() ? "Paste as rich text"
-                                                      : "Paste as plain text"),
+                         pastePlaintextByDefault() ? tr("Paste as rich text")
+                                                   : tr("Paste as plain text"),
                          this);
     action->setShortcut(AlternatePasteShortcut);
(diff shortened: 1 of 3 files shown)


client/chatedit.cpp
```diff
@@ -70,6 +76,22 @@ void ChatEdit::contextMenuEvent(QContextMenuEvent *event)
     connect(action, &QAction::triggered, this, &KChatEdit::resetCurrentFormat);
     menu->addAction(action);
 
+    action = new QAction(QIcon::fromTheme("edit-paste"),
+                         tr(pastePlaintextByDefault() ? "Paste as rich text"
+                                                      : "Paste as plain text"),
```
KitsuneRal: I don't think this works with Qt Linguist - it needs both strings to be there in separate, so that they land in separate translation slots.
```suggestion
                         pastePlaintextByDefault() ? tr("Paste as rich text")
                                                   : tr("Paste as plain text"),
```

berrange:
While this looks like as a usable workaround, I still feel like the current paste-as-HTML impl is fundamentally broken in its implementation. If it can't reliably paste in HTML that has been selected in the browser window, it does not look fit for purpose as is. Pasting 100% complete well formed HTML documents does not look likely to be the common case.

KitsuneRal (changes requested):
Thanks - overall LGTM, save for one fix.

KitsuneRal (approved):

PR #855, opened 2023-02-22
RoomListModel: refresh room if highlights changed

rpallai:
`Room::Change::Highlights` was introduced in Quotient-0.7 and must be captured here because it changes model data.


KitsuneRal (approved):
Thanks!

PR #858, opened 2023-03-30
Make login more robust against faulty cases

### LINKED ISSUE #830: Login duplicates
### Description

After each new launch of Quaternion there is a line about the new login in the list of recent logins (which I look in app.element.io). There are 85 of them now.
The password is saved in Quaternion. Quaternion is used by my brother (on Windows). There is no such thing in my Linux build.

### Steps to reproduce

- Launch Quatrenion on Windows
- Chatted
- Closed Quatrenion
- Repeated this several times, every day

- Opened the browser
- Logged in to app.element.io
- Opened Security section
- There are many duplicates of Quatrenion in the list of recent (active) logins
- I delete the old ones, but it repeats

I expect there should be one login. Like in the Linux version.

### Version information

- **Quaternion version**: 0.0.95.1
- **Install method**: archive downloaded from this site https://github.com/quotient-im/Quaternion/releases + OpenSSL from here https://slproweb.com/products/Win32OpenSSL.html Win64 Installer Light. I don't remember the version, I think it's 1.1.1
- **Platform**: Windows 7 SP1 x64 with all updates.
- Private chat (room for two of us) without E2E encryption, matrix.org server.

KitsuneRal:
Main part of that is pulling the login code back from `AccountRegistry` into `MainWindow`.

### DIFF
--- a/client/accountselector.cpp
+++ b/client/accountselector.cpp
@@ -1,38 +1,43 @@
 #include "accountselector.h"
 
+#include <accountregistry.h>
 #include <connection.h>
 
 using namespace Quotient;
 
-AccountSelector::AccountSelector(QWidget *parent)
+AccountSelector::AccountSelector(AccountRegistry* registry, QWidget* parent)
     : QComboBox(parent)
 {
+    Q_ASSERT(registry != nullptr);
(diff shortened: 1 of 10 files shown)


KitsuneRal:
There were unrelated commits before the force-push, disregard them.

PR #862, opened 2023-04-05
readme: remove lgtm badge

circumspect:
LGTM no longer exists, and the badge also says "no longer available"
https://github.blog/2022-08-15-the-next-step-for-lgtm-com-github-code-scanning/


KitsuneRal:
Indeed, and we don't really use LGTM since autumn.

PR #863, opened 2023-04-08
Port to SPDX

### LINKED ISSUE #786: Switch to SPDX

TobiasFella:
Done using https://invent.kde.org/sdk/licensedigger with the quaternion-style GPL-3.0-or-later header patched in

Some "boxes" were then fixed manually to not be entirely broken; the style of the comments is still not consistent but that's not caused by this MR


CarlSchwan (approved):

.clang-format
```diff
@@ -1,3 +1,5 @@
+# SPDX-FileCopyrightText: 2019 Project Quotient
+# SPDX-License-Identifier: LGPL-2.1-only
 # Copyright (C) 2019 Project Quotient
```
KitsuneRal:
```suggestion
```

KitsuneRal (approved):
Thanks for the conversion!

PR #868, opened 2023-05-04
Check for image MIME data before HTML data

### LINKED ISSUE #867: Pasting an image from a browser to the chat window crashes on send
### Description

Copy&Pasting an image from Firefox looks to paste something in the textbox, but upon sending Quaternion crashes with an assert failure: ASSERT: "!plainText.isEmpty() && !htmlText.isEmpty()" in file /.../client/chatroomwidget.cpp, line 428

### Steps to reproduce

- Copy an image, test from a web page
- Paste in Quaternion chat field
- Hit ENTER to send

Ideal behavior would be to convert the paste into an image upload, like Element does.  However, not crashing, or even just outright rejecting an image paste, would also work.

Adding some printf debugging to ChatEdit::insertFromMimeData(), I see the following:

hasHtml:  true  hasImage:  true 

html:  "<meta http-equiv=\"content-type\" content=\"text/html; charset=utf-8\">"  -- end

"html: " and " -- end" being added by me as delimiters.

Quaternion dumps logs to the standard output. If you can find the logs and
identify any log snippets relevant to your issue, please include
those here (please be careful to remove any personal or private data):

ASSERT: "!plainText.isEmpty() && !htmlText.isEmpty()" in file /.../client/chatroomwidget.cpp, line 428

### Version information

- **Quaternion version**: 0.0.96-beta
- **Qt version**: 5.15.8
- **Install method**: Build from source (zip file download from GitHub) 
- **Platform**: Debian Bookworm, ppc64el

chmeeedalf:
Copying an image from Firefox attaches some HTML to the image. Extracting the plain text from the HTML before the image results in empty data in the field, with no image, triggering an assertion failure on message send.

This is a partial fix for issue #867.  Partial, because the resulting image does not appear in the chat window until relaunching, due to an error opening the temp file created by this.

### DIFF
--- a/client/chatedit.cpp
+++ b/client/chatedit.cpp
@@ -110,7 +110,9 @@ void ChatEdit::insertFromMimeData(const QMimeData *source)
         return;
     }
 
-    if (source->hasHtml()) {
+    if (source->hasImage())
+        emit insertImageRequested(source->imageData().value<QImage>());
+    else if (source->hasHtml()) {
         if (m_pastePlaintext) {
             QTextDocument document;
             document.setHtml(source->html());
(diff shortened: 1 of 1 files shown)


KitsuneRal (approved):
Thanks! Makes perfect sense.

PR #870, opened 2023-05-12
Haiku: Don't check for dbus

ghost:


KitsuneRal:
Thanks for getting my attention to this. It seems that the DBus dependency is just not needed after all, even on Linux systems. I will kick it out separately.

Issue #845, opened 2023-01-24
How do I add a contact?


narcisgarcia:
- Debian GNU/Linux 11/Stable (amd64) +Gnome
- Quaternion from OS package 0.0.9.4c-1build1

I don't find any option to "Add a contact" or invite some individual (who's ID I already know) to chat with me.
Please document and illustrate steps.

KitsuneRal:
The documentation on `/`-commands is massively overdue indeed... To quickly help - you can just type `/join @user:server` in the message box (in any room or without a room opened at all).

narcisgarcia:
For every command there should be some GUI button or menu option.

narcisgarcia:
(of course it's @myfriend:example.net)
`/join @user:server`

Status bar says:
/join argument doesn't look like a room ID or alias

KitsuneRal:
Sorry, I mistook the command indeed: `/join` is used to join an existing room, and to message a person you should use `/msg` or `/pm`.

PR #871, opened 2023-05-17
add graphical option to add contacts (create private chats)

MoKe12g:
Adds an option to the menu to create a private chat as wished in #845.


KitsuneRal (changes requested):
Thanks, and sorry for taking time to review. The existing `Open room...` dialog supports entering user ids too. Besides, `Open room...` supports `matrix:` URIs and `matrix.to` URLs. I recognise it's not quite discoverable - happy to get suggestions how to improve that. I can understand that having a separate dialog restricted to user ids might be the best solution  - but then please extend `openUserInput()` instead of duplicating most of its code. Turning a `bool` parameter to an `enum` and having two or three extra conditionals is totally fine for that matter.

PR #895, opened 2024-01-31
[fix] close file on each try to check attachment

mizhka:
It fixes error "already-open-file" due to multiple tries to check attachment.


mizhka:
Overcome by events.
Thanks to @KitsuneRal  for quick fix!

KitsuneRal:
Thanks for reporting in the first place :)

PR #907, opened 2024-10-13
Added a confirmation message box before forgetting a room

### LINKED ISSUE #236: Forgetting should be confirmed by the user

dymoksc:
Fixes #236

### DIFF
--- a/client/roomlistdock.cpp
+++ b/client/roomlistdock.cpp
@@ -11,6 +11,7 @@
 #include "logging_categories.h"
 
 #include <QtWidgets/QMenu>
+#include <QtWidgets/QMessageBox>
 #include <QtWidgets/QStyledItemDelegate>
 #include <QtWidgets/QLabel>
 #include <QtWidgets/QPlainTextEdit>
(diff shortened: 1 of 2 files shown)


client/roomlistdock.cpp
```diff
@@ -194,10 +195,14 @@ RoomListDock::RoomListDock(MainWindow* parent)
...
         roomContextMenu->addAction(QIcon::fromTheme("irc-remove-operator"),
         tr("Forget room"), this, [this] {
-            if (auto room = getSelectedRoom())
-            {
-                Q_ASSERT(room->connection());
-                room->connection()->forgetRoom(room->id());
+            if (auto room = getSelectedRoom()) {
+                QMessageBox::StandardButton confirmation = QMessageBox::question(
+                    this, tr("Forget this room?"),
+                    tr("Are you sure you want to forget room %1?").arg(room->name()));
```
```suggestion
                    tr("Are you sure you want to forget room %1?").arg(room->displayName()));
```
`room->name()` may return an empty string if the room has no set name; `displayName()`, on the other hand, tries really hard to produce something human-friendly, trying the room name, canonical alias, and even listing a few members if everything else fails.

client/roomlistdock.cpp
```diff
@@ -194,10 +195,14 @@ RoomListDock::RoomListDock(MainWindow* parent)
...
+                if (confirmation == QMessageBox::Yes) {
+                    Q_ASSERT(room->connection());
+                    room->connection()->forgetRoom(room->id());
```
```suggestion
                    if (QUO_CHECK(room->connection())
                        room->connection()->forgetRoom(room->id());
```
We have a safer alternative to `Q_ASSERT` in libQuotient.

KitsuneRal (approved):
Many thanks for the contribution! Looks good, aside from a couple of suboptimal places - I will fix them myself shortly.

PR #908, opened 2024-10-13
Remove users to invite from list

### LINKED ISSUE #650: Create room dialog does not allow to remove people to invite
### Description

When creating a new room in Quaternion, users can only be added to the list. A mistake cannot be undone, there is no option to remove users from the list (but they have not yet been really invited).

**Quaternion version**: commit f6e376414b1c69f32d2437f1b60ee1e3a175d968

dymoksc:
Fixes #650


KitsuneRal (approved):
Thanks for the contribution! In the future, you generally shouldn't add `quaternion_en.ts` to pull requests as it gets updated very frequently and can easily lead to merge conflicts (as happened in this case).

PR #946, opened 2026-03-17
Join room via link

### LINKED ISSUE #942: [Bug] Invalid matrix (hash) links
Clicking a link to a matrix room from another room's description fails to invoke navigation / redirect ~~due to what appears to be a malformed string error.~~

For instance the following link will fail to redirect:
Image

An alert will pop up showing the target href.
Image

This is accompanied by the following log:
```
quotient.main: Room for alias "#<roomname>:matrix.org" is not found under account "@<username>:matrix.org"

```

ford-jones:
Resolves #942 

* When clicking a matrix link that is _not_ currently in the users room list, display a popup prompting the user with the option to join the room.

* Removes build artifacts from check-in

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ flatpak/.flatpak-builder
 flatpak/repo
 
 CMakeCache.txt
+CMakeFiles/
 cmake_install.cmake
 Makefile
 quaternion_autogen/
@@ -20,3 +21,10 @@ quaternion_autogen/
 .cache
 
 .DS_Store
+
+quaternion
+*.qm
+
+.qt
+.lupdate
+.vscode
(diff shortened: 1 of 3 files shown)


ford-jones:
Does this project have a linter configuration (e.g. trunk) ? Apologies if this does not meet the defined code-style. I'm happy to make changes if this is a problem.

KitsuneRal:
We use ClangFormat and ClangTidy for linting (configuration files are in the tree); also, as you can see above, Sonar analyses the PRs (since not long ago, actually - a lot of what it finds, in the PR or elsewhere, can be either a false positive, or an easy target for fixing).

ford-jones:
`ClangFormat` came up clean but I removed the whitespace that sonarqubecloud flagged as an issue. Should be good to go :+1:

ford-jones:
Anything to do here @KitsuneRal ?

KitsuneRal:
No, it just needed my time - sorry for the wait. That looks quite reasonable for starters. Perhaps eventually we could add a preview of the room if it allows non-member access but this is already much better than what it used to be.

PR #947, opened 2026-03-18
Render message body alongside file content

### LINKED ISSUE #587: Message body is suppressed when a file is attached
Quaternion allows to enter the message body along with the file; but then it only shows the filename rather than the body. It may make sense to show both in some reasonable way, if they don't duplicate each other and are non-empty.

ford-jones:
Resolves #587 

* Ensures that the text content included in messages that contain file attachments are rendered alongside the filename of the attached media

### DIFF
--- a/client/models/messageeventmodel.cpp
+++ b/client/models/messageeventmodel.cpp
@@ -497,9 +497,14 @@ QString MessageEventModel::visualiseEvent(const Quotient::RoomEvent& evt, bool a
                 return cleanHtml;
             }
             if (const auto fileContent = e.get<FileContentBase>()) {
-                auto fileCaption = fileContent->commonInfo().originalName.toHtmlEscaped();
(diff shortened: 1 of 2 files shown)


KitsuneRal:
This is probably fine but I'm not sure how much showing the file name is valuable when a proper caption is there. One will see the suggested filename when downloading the file anyway. I don't have an opinion. If you have considerations, please share.

ford-jones:
@KitsuneRal I agree -- I don't think it's entirely necessary, I only appended the two together because #587 says:
> It may make sense to show both in some reasonable way

(granted this ticket is ~7 years old :sweat_smile:)

I don't have strong feelings about it either way, so I'm happy to go with whatever you think is best here!

KitsuneRal:
As discussed in the chat - let's have it the way it's done, and we'll see how it goes.

PR #948, opened 2026-03-29
Redaction confirmation

### LINKED ISSUE #822: `Redact` confirmation dialog that shows the message
It is too easy to redact the wrong line.

- Redact should ask for confirmation
- The question should contain the target message

ford-jones:
Resolves #822 

* Shows a `QMessageBox` dialogue prompting the user to confirm a redaction
* Displays a preview of the text that is being removed
* When an image is selected for redaction the filename is shown
* Selecting `Yes` redacts the event, `No` cancels the action

### DIFF
--- a/client/timelinewidget.cpp
+++ b/client/timelinewidget.cpp
@@ -188,13 +188,13 @@ void TimelineWidget::showMenu(int index, const QString& hoveredLink,
     {
         auto plainText = modelIndex.data().toString();
         menu->addAction(QIcon::fromTheme("edit-delete"), tr("Redact"), this,
-                        [this, eventId, plainText] { 
+                        [this, eventId, plainText] {
                             QMessageBox::StandardButton join = QMessageBox::question(this, tr("Confirm redaction"),
                                 tr("Are you sure you want to redact the following message?: %1")
                                 .arg("\"" + plainText + "\""));
-                
+
                             if(join == QMessageBox::Yes)
-                                currentRoom()->redactEvent(eventId); 
+                                currentRoom()->redactEvent(eventId);
                         });
     }
 
--- a/client/timelinewidget.cpp
+++ b/client/timelinewidget.cpp
@@ -4,25 +4,22 @@
 #include "logging_categories.h"
 #include "models/messageeventmodel.h"
 
+#include <Quotient/csapi/message_pagination.h>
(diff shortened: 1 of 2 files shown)


ford-jones:
@KitsuneRal The issue flagged by sonar here has been present since long before this PR. It's just only now been picked up due to `clang-format` in-lining the booleans.

KitsuneRal:
Yeah I know - Sonar has that habit of pointing out old issues with a rationale being "you can fix this too while you're at it". Not that I always agree, and the one it flagged this time I don't quite agree in general - this particular code hasn't changed since it was written and I wouldn't touch it just for the sake of fixing a warning.

client/timelinewidget.cpp
```diff
@@ -173,94 +165,88 @@ void TimelineWidget::onMessageShownChanged(int visualIndex, bool shown,
...
-                        [this, eventId] { currentRoom()->redactEvent(eventId); });
+                        [this, eventId, plainText]
+        {
+            QMessageBox::StandardButton join =
+              QMessageBox::question(this, tr("Confirm redaction"),
+                                    tr("Are you sure you want to redact the following message?: \"%1\"")
+                                      .arg(plainText));
+
+            if (join == QMessageBox::Yes)
+                currentRoom()->redactEvent(eventId);
```
KitsuneRal:
```suggestion
            if (QMessageBox::question(this, tr("Confirm redaction"),
                                      tr("Are you sure you want to redact the following message?: \"%1\"")
                                        .arg(plainText.left(100)))
                 == QMessageBox::Yes)
                currentRoom()->redactEvent(eventId);
```
Aside from eliminating a variable that is only used once - some people can be quite verbose and put a lot of text in a single message. As the most naive way around this, I suggested using `left()`; but feel free to come up with something more intelligent (cutting the message at a non-word character, e.g.).

KitsuneRal (changes requested):

client/timelinewidget.cpp
ford-jones: Oh I hadn't even considered this! Good idea, I'm happy with the changes you've made here :)

KitsuneRal (approved):

KitsuneRal:
Thanks a lot!

